PR #20, opened 2026-04-09
chore: update license headers for consistency

privateip:
The project has inconsistencies in the source code with some files displaying an Apache source header while the repository (project) is
licensed AGPL.   The Python source code was missing source code headers.

All source code is now properly displaying the AGPL header and copyright.

Additionally, a new `make` target was added to use `go-licenses` to generate the `NOTICE` file.  The `NOTICE` file contains copies of all license attributions for libaries used in this project.

To generate the `NOTICE` file, simply run `make notice` from the root of the repository.


drewr (approved):
Good catch 👍

PR #23, opened 2026-04-17
chore: clean up tasks from repo consolidation

privateip:
When the Galactic repos were merged into a single monorepo, there were some outstanding touch points that were still cross repository.  This patch cleans those up to use only the monorepo `galactic`.

The `operator.k8s.yml` file was being pulled from the archived `galactic-operator` repo.  The manifest has been copied from the `galactic-operator` repo to this repository and referened locally in the `Dockerfile` when building the Kind node.

The `install.sh` script has been updated to use the local `operator.k8s.yml` manifest file instead of pulling it from the `galactic-operator` repo.

The `Dockerfile` that is used to build the Kind node has been updated to include `tcpdump` for troubleshooting.

### DIFF
--- a/lab/kindest-node-galactic/install.sh
+++ b/lab/kindest-node-galactic/install.sh
@@ -34,8 +34,8 @@ if hostname |grep -q control-plane; then # control-plane
   kubectl -n galactic-mqtt rollout status deployment galactic-mqtt
 
   # Galactic Operator
-  curl -L "https://raw.githubusercontent.com/datum-cloud/galactic-operator/refs/heads/main/dist/install.yaml" |sed -e "s/galactic:latest/galactic:${GALACTIC_VERSION}/g" |kubectl apply -f -
-  kubectl -n galactic-operator-system rollout status deployment galactic-operator-controller-manager
+  kubectl apply -f /galactic/operator.k8s.yml
+  kubectl -n galactic-system rollout status deployment galactic-controller-manager
 
   # Galactic Router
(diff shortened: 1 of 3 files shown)


lab/kindest-node-galactic/install.sh
```diff
@@ -34,8 +34,8 @@ if hostname |grep -q control-plane; then # control-plane
   kubectl -n galactic-mqtt rollout status deployment galactic-mqtt
 
   # Galactic Operator
-  curl -L "https://raw.githubusercontent.com/datum-cloud/galactic-operator/refs/heads/main/dist/install.yaml" |sed -e "s/galactic:latest/galactic:${GALACTIC_VERSION}/g" |kubectl apply -f -
-  kubectl -n galactic-operator-system rollout status deployment galactic-operator-controller-manager
+  kubectl apply -f /galactic/operator.k8s.yml
```
scotwells, line 37: Should this be an absolute path?

lab/kindest-node-galactic/operator.k8s.yaml
scotwells, line 1: Seems like this largely duplicates what's already in [`config`](https://github.com/datum-cloud/galactic/tree/main/config)?

scotwells (approved):

lab/kindest-node-galactic/install.sh
privateip: yes, its invoked in the container

lab/kindest-node-galactic/operator.k8s.yaml
privateip: Could very well be the case.  This PR was just attempting to replicate exactly how things where configured before the repo consolidation.   I suspect there are a number of things that can and should get cleaned up yet.

PR #70, opened 2026-05-06
chore: configure renovate with grouped dependencies and automerge rules

privateip:
Set up Renovate with dependency grouping for Go modules and GitHub Actions, monthly cadence for k8s core libraries, automerge for patch/minor non-k8s deps, vulnerability alerts, and weekly scheduling.


scotwells (approved):
🙏

PR #74, opened 2026-05-08
feat: Replace MQTT control plane with embedded BGP

privateip:
Galactic's distribution path moves from the Python galactic-router service over MQTT to standard MP-BGP L3VPN with SRv6 service SIDs. Each agent now embeds a GoBGP speaker peering iBGP with one or more cluster route reflectors; the router service and its broker are removed entirely.

Design rationale and decision log: PLAN-bgp-cutover.md.
Component-level reference for the resulting system: AS-BUILT.md.

Major changes:

- VPC identifiers narrow from 48 to 32 bits so the BGP route distinguisher (RFC 4364 type 2) and route target extended community (RFC 5668) carry the full ID without truncation. Attachment IDs remain 16 bits.
- The operator gains --pop-locator and --asn flags; reconciler writes serviceSID, routeTarget, and routeDistinguisher into VPCAttachment.status alongside the identifier.
- VPCAttachmentSpec.Routes and the derived NAD terminations are removed; pod prefixes are the only thing advertised. Anything reachable beyond the pod is learned via BGP.
- New agent packages: bgp (GoBGP wrapper, originate/withdraw/watch, tunnel-encap and prefix-sid encodings with a startup-time round-trip guard), cache (joined VPC + VPCAttachment informer), program (kernel egress with explicit hex/base62 boundary), reconciler (central event loop with pending/active/ingress/ programmedKernel/receivedRoutes state), health (/healthz + /metrics), metrics (Prometheus gauges).
- Deleted: pkg/proto/remote, internal/agent/srv6/{srv6.go, routeegress, neighborproxy}, internal/cni/route, the entire router/ Python service, the lab MQTT/router manifests, and the paho.mqtt.golang dependency.
- Cluster-wide configuration (POP-locator, ASN, BGP encoding) flows to agents via the operator ConfigMap so wire format and authority cannot diverge per-node.

Tests, lint, and code generation are clean. Reconciler ordering and teardown paths have unit coverage; the prefix-sid wire format has a round-trip regression guard against the pinned GoBGP version.


scotwells:
Should we be tying galactic to GoBGP's api or have it integrate with a more generic BGP control plane so we can use a standard control plane across BGP?

privateip:
Right now GoBGP is really the only game in town unless we roll our own which we aren't going to do or want to move to a sidecar architecture which i would rather we didn't.   That said, we could abstract the GoBGP implementation into a more abstract interface making it more pluggable for future considerations.

scotwells:
Agree we should use GoBGP. I was just brainstorming about the possibility of abstracting how we configure it through the k8s control plane. That's what inspired the work in https://github.com/milo-os/bgp

privateip:
~ah gotcha, im 100% in agreement.  I needed to put a stake in the ground and cut this PR off before it got any bigger.  Its way to big as it is.   I will update [#594](https://github.com/datum-cloud/enhancements/issues/594) with the details of done but would like to get this one in so we can work towards more pointed PRs that aren't trying to boil the ocean.~ :facepalm:

privateip:
@scotwells Ok, after thinking about this some more I think i missed the point of your comment.  Are you thinking that Milo will reconcile the BGP CRD and therefore Galactic doesn't need to. It only needs to provide APIs?

scotwells:
@privateip yes, I'd expect the BGP controller to reconcile resources in the Milo control plane and program GoBGP. Galactic would be able to just configure BGP resources in Milo. I figured we could use the same BGP control plane for the rest of our infrastructure as well.

privateip:
Now we are on the same page :facepalm: my bad.    I'm going to go back and re-work this PR a bit, stay tuned

privateip:
closing this PR, will re-open a different PR that addresses this

PR #84, opened 2026-05-22
chore: Remove unused GVPC lab node config files

privateip:
These per-node files (frr.conf, gobgp.conf, startup.sh, daemons) were superseded when the lab was updated to use the official GoBGP image and ContainerLab-managed startup configuration in PR #82. Removing them to avoid confusion over which configs are actually in use.


scotwells (approved):
Clean up time!

PR #96, opened 2026-05-25
ci: enforce YAML linting, proto drift detection, and pre-commit hooks

0xmc:
- Add yaml-lint CI job backed by .yamllint.yaml (120-char line limit, document-start disabled for Taskfiles/GH Actions, 'on' key allowed)
- Add proto-check CI job: regenerates .pb.go files and fails on diff, preventing committed proto drift; gated on test-e2e alongside other tier-1 jobs
- Expand unit test coverage from ./pkg/common/util/... to ./... with -short flag so network-dependent tests are skipped in CI
- Add .pre-commit-config.yaml: trailing-whitespace, end-of-file-fixer, merge-conflict detection, private-key detection, yamllint, gofmt, go vet, and golangci-lint
- Add task hooks (installs pre-commit hooks) and task proto (regenerates protobuf Go code); pin all toolchain versions: golangci-lint v2.12.2, protoc 29.3, protoc-gen-go v1.36.11, protoc-gen-go-grpc v1.6.1
- Update both devcontainers to install Python 3.13 (pre-commit runtime), run task hooks on post-create, and use pinned tool versions matching CI


.github/workflows/ci.yml
```diff
@@ -89,11 +101,36 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
+  proto-check:
```
privateip, line 104: The proto implementation is being phased out of the code base so this is not needed going forward.

.github/workflows/ci.yml
```diff
@@ -27,6 +27,18 @@ jobs:
           version: v2.12.2
           args: --timeout=5m
 
+  yaml-lint:
```
privateip, line 30: If we are going `yamllint` here, it makes the pre-commit check somewhat redundant.   I would rather we do it in the CI pipeline and not do using the pre-commit hook.  Happy to discuss if you feel strongly about using pre-commit

privateip (changes requested):

.github/workflows/ci.yml
0xmc: I'll get this pulled out

.github/workflows/ci.yml
0xmc: I generally like the idea of catching these types of things with a pre-commit hook or something I can run locally before CI catches them.  Ideally it matches what CI is doing, which granted is duplicative, but I like to think of it as hygiene in depth.

privateip: Agreed with the localization of it for sure.   I have a todo item on my list to pull the execution out of the pipeline and into localized scripts so the pre-merge ci actions can be run locally.   would that work address your concern?

0xmc:
Absolutely.  Lately I've been adding a `quality` target to my Make^w
Taskfiles so I can continually run these checks while developing, not just
at commit time.  Establishing uniform quality gates across the SDLC would
be the bee's knees.

0xmc:
With your pre-merge ci actions, I'm just going to close the PR and make any future updates to that tooling.

PR #116, opened 2026-06-17
feat(gobgp): expose BGPProviderService on galactic's gRPC port

privateip:
## Summary

- Replaces the separate GoBGP gRPC API port with a `BGPProviderService` registered on galactic's own health gRPC server (`--grpc-health-port`). Cosmos now configures GoBGP by calling galactic's `BGPProviderService` rather than dialing GoBGP's API directly, eliminating `--gobgp-api-port` and the extra per-node port.
- Removes `BGPProvider` deletion on shutdown — provider lifecycle is now managed by cosmos.
- Switches `BGPProviderSpec.Endpoint` to a top-level field (drops the `GoBGP` nested config struct) and bumps the cosmos dependency to pick up the updated API.

## Changes

- `internal/gobgp/provider.go` — new `BGPProviderServiceServer` implementation
- `internal/agent/agent.go` — registers `BGPProviderService` on the shared `grpcSrv`; passes `grpcAddr` to bootstrap instead of `gobgpSrv.Addr()`
- `internal/gobgp/server.go` — removes `APIPort`/`Addr()`; switches to `atomic.Pointer[BgpServer]` + channel-based `ready` signal; adds `Reconfigure()` for future reconfiguration support
- `internal/bootstrap/bootstrap.go` — updates `BGPProviderSpec` to new flat `Endpoint` field
- Deployment manifests — removes `--gobgp-api-port` flag from all daemonset specs
- `deploy/containerlab/resources/cosmos/` — adds kustomize overlay for Kind (replaces manual kubectl patch in `install-overlay.sh`)
- `go.mod/go.sum` — bumps `go.miloapis.com/cosmos` to `20260616`

## Test plan

- [ ] `task lint` passes
- [ ] `task test` passes (unit + e2e)
- [ ] ContainerLab: `deploy_cosmos` deploys via kustomize overlay without manual patches
- [ ] Agent starts with `--gobgp-enabled=true`; cosmos configures GoBGP via `BGPProviderService` on the health port
- [ ] Kubernetes readiness probe reports SERVING after GoBGP initializes


cmd/galactic-agent/main.go
```diff
@@ -37,12 +37,10 @@ func newRootCommand() *cobra.Command {
 		"BGP plane label published on the BGPProvider (e.g. overlay, overlay-rr)")
 	cmd.Flags().BoolVar(&opts.GoBGPEnabled, "gobgp-enabled", false,
 		"Enable embedded GoBGP and publish BGPProvider")
-	cmd.Flags().IntVar(&opts.GoBGPAPIPort, "gobgp-api-port", 50051,
-		"Port for the embedded GoBGP gRPC API (cosmos dials this)")
 	cmd.Flags().StringVar(&opts.GoBGPLogLevel, "gobgp-log-level", "panic",
 		"GoBGP internal log level (debug, info, warn, error, panic)")
 	cmd.Flags().IntVar(&opts.GRPCHealthPort, "grpc-health-port", 8084,
```
scotwells: The name of this poet is a little off now. Also might be worth having these on separate ports so we can have different authentication?

privateip: ugh thought i had caught all of these.  Its actually actually the `galactic-agent` port not GoBGP which is entirely abstracted and embedded in `galactic-agent`.  Correction coming.

scotwells (approved):

PR #117, opened 2026-06-17
refactor(cni): rename cmd/galactic to cmd/galactic-cni and simplify to pure CNI plugin

privateip:
## Summary

- Renames `cmd/galactic/` to `cmd/galactic-cni/` to match the binary name
- Strips the entire cobra CLI layer from `galactic-cni`: the binary is always invoked by the container runtime with `CNI_COMMAND` set, so the root command, `cni`/`version` subcommands, and env-var dispatch were dead code
- `main()` is now three lines — import and call `cni.RunPlugin()`
- Updates CNI `"type"` from `"galactic"` to `"galactic-cni"` in all NADs and test fixtures to match the installed binary name
- Removes `internal/cmd/cni` (a pass-through cobra wrapper with no purpose) and `NewCommand()` from `internal/cmd/version`

## Test plan

- [ ] `task build` produces `bin/galactic-cni` and `file` confirms it's an ELF binary
- [ ] `go test ./internal/cni/...` passes
- [ ] `go build ./...` clean
- [ ] Verify NAD `"type": "galactic-cni"` matches the binary installed by `install.sh` at `/opt/cni/bin/galactic-cni`


scotwells (approved):

scotwells:
Should we consider having CNI and agent be sub commands of the galactic binary instead? Then we'd only need to build a single binary artifact and release

privateip:
We could combine this into a single binary however CNI doesn't invoke subcommands so you end up bypassing your command CLI command loop anyway.  You end up with an implementation that looks something like:

```go
func main() {
	if os.Getenv("CNI_COMMAND") != "" {
                // Run CNI loop
        }

	// Normal CLI mode with subcommands
	rootCmd := &cobra.Command{
                // Normal CLI loop with subcommands
        }
}
```

So the code paths diverge the moment the binary is run which feels *off* to me.

PR #131, opened 2026-06-25
feat: RBAC pre-flight check, containerlab fixes, controller refactor

privateip:
## Summary

This is a set of patches to fix the containerlab deployment to work with the latest CRDs published by Cosmos.   It also adds a new function to log RBAC failures before attempting to start `galactic-router`.   Prior to this patch, `galatic-router` would look like it starts fine but never established peer connections.

## Changes

- **RBAC pre-flight check** — `checkWatchPermissions()` issues a test list for every resource the manager watches at startup. If any return Forbidden, logs a clear actionable message so cache sync failures are immediately visible rather than silent.
- **Containerlab IPv6 fixes** — Skip TLS verify in kubectl wrapper for IPv6-only kind clusters; add `lo-galactic` dummy interface with IPv6 address for transit mesh routing.
- **RBAC migration** — Rename from `galactic-agent` → `galactic-router`; update apiGroups from `providers.bgp.miloapis.com` → `bgp.miloapis.com`; add permissions for all new CRDs (BGPRouter, BGPPeer, BGPAdvertisement, BGPPolicy, BGPVRFInstance) plus Node and Secret.
- **Containerlab routing** — Add `BGP_LOCAL_ADDRESS` env for iad-rr tenant; add FRR route-map SET_SRC to set src address for EVPN routes.
- **Remove ASN patch** — `fix-asn-maximum.sh` no longer needed after cosmos API update.
- **Cosmos API update** — `Prefix` type changed from `AdvertisedPrefix.CIDR` to direct `Prefix` string; added int32→uint32 conversion for LOCAL_PREF (cosmos uses int32, GoBGP uses uint32 per RFC 4271).
- **Controller refactor** — Moved `BGPAdvertisement` Watches handler into `BGPRouter.SetupWithManager`; removed redundant empty reconciler.
- **Fix field index names** — Corrected `BGPPeerByRouterName` → `BGPAdvByRouterName` and `BGPPeerByRouterName` → `BGPPolicyByRouterName` in status update list calls.


cmd/galactic-router/main.go
```diff
@@ -174,3 +187,39 @@ func main() {
...
+		&corev1.NodeList{},
+	}
+
+	for _, objList := range watched {
+		if err := c.List(ctx, objList, client.Limit(1)); err != nil {
+			if apierrors.IsForbidden(err) {
+				logger.Error(err, "missing list+watch RBAC",
+					"detail", "informer cache will not sync; add resource to ServiceAccount ClusterRole and restart")
+			}
+		}
```
scotwells: I'd recommend doing a SelfSubjectAccessReview check here. It's a simple permission check instead having to list resources. You can also check for the Watch permission when this is only checking the list permission.

scotwells (dismissed):

scotwells:
Assume we should have also have alerts that catch these situations? E.g. the controller is erroring when making API requests

privateip:
> Assume we should have also have alerts that catch these situations?

yes, i have held off putting alerts into galactic until the code settled down. Im very close to starting that work now across the board

scotwells (approved):

PR #132, opened 2026-06-25
feat: replace SRv6 endpoint routing with IPAM-managed pod subnet alloc

privateip:
## Summary

This patch adds a new resource that can be used to test and exercise the
CNI connection.   The once the containerlab deployment is active running
`task deploy:testvpc` will deploy pods to each cluster and craete a VPC
between them

## Changes

- Add internal/cni/ipam package with PoolAllocator and StaticAllocator for IPv6 /80 subnet allocation from a CIDR pool
- Rewrite cmdAdd to configure IPAM, set IP address and default route on the guest interface inside the container netns, and advertise the allocated pod subnet in the BGPAdvertisement CRD
- Rewrite cmdDel to deallocate IPAM by reading the allocated subnet from the BGPAdvertisement CRD annotation keyed by container ID
- Remove SRv6 ingress route setup from CNI; the controller now derives the SRv6 endpoint from the pod subnet and installs the ingress route
- Add idempotency guards to veth.Add and vrf.Add so a leftover interface from a prior failed cmdAdd does not cause cmdAdd to fail
- Wrap galactic-cni binary with a shell script that exports NODE_NAME and KUBECONFIG so the CNI binary can reach the API server in Kind
- Add containerlab test VPC resources (NAD, nginx deployment, RBAC) and install-testvpc.sh script to deploy cross-cluster connectivity
- Create /var/run/galactic-cni directory in Dockerfile for IPAM state
- Update BGPVRFInstance route target format to drop "rt:" prefix

### DIFF
--- a/containers/galactic/Dockerfile
+++ b/containers/galactic/Dockerfile
@@ -28,11 +28,17 @@ RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build \
       -X go.datum.net/galactic/internal/metadata.BuildDate=${BUILD_DATE}" \
     -o galactic-cni cmd/galactic-cni/main.go
 
+# Create the IPAM state directory in the builder (where /bin/sh exists)
+# so it can be copied into the final distroless image.
+RUN mkdir -p /var/run/galactic-cni && chmod 777 /var/run/galactic-cni
+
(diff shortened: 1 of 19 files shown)


scotwells (approved):
~Assume this IPAM CNI is just for testing?~ Ignore me! I just read the description

PR #137, opened 2026-06-26
bugfix: implement SRv6 EVPN data plane for inter-node VPC routing

privateip:
## Summary

The local IPAM implement in the CNI exposed a bug with how SRv6 routes were being programmed for the data plane.  The routes are now properly programmed with the correct next hop address.

## Changes

- Encode End.DT46 SID in galactic-cni and write it to the BGPAdvertisement annotation so galactic-router can advertise the correct seg6 encap target
- Place SRv6SID in EVPN Type 5 GWIPAddress instead of the BGP peering loopback so receiving nodes install SEG6 encap routes targeting the actual decap SID
- Add watchEVPNRIB goroutine that installs and removes kernel SEG6 encap routes into the VRF routing table as EVPN best paths are added or withdrawn
- Add probeRouteWrite startup check that fails fast with a clear hint when the container lacks root or CAP_NET_ADMIN, before the first EVPN event arrives
- Fix subnet-router anycast absorption  by using /128 for the host veth gateway address and explicitly installing the pod subnet route into the VRF
- Add runAsUser: 0 to containerlab overlay and production DaemonSet manifests; fix Dockerfile to build the full cmd/galactic-router package directory
- Thread BGP_LOCAL_ADDRESS through Reconciler as the EVPN next-hop override for environments where the node InternalIP is not transit-reachable

### DIFF
--- a/cmd/galactic-router/root.go
+++ b/cmd/galactic-router/root.go
@@ -223,7 +223,7 @@ func runCmd(v *viper.Viper) error {
 	runtimeMgr := galacticruntime.NewRuntimeManager(factory)
 
 	// Create reconciler.
-	rec := reconcile.New(mgr.GetClient(), nodeName, routerRole)
+	rec := reconcile.New(mgr.GetClient(), nodeName, routerRole, bgpLocalAddr)
 
 	// Register BGPRouter controller.
 	if err := (&controller.BGPRouterReconciler{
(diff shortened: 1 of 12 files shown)


ecv:
trailing slash or no @privateip

scotwells (approved):

PR #149, opened 2026-06-30
feat: add garbage collector for orphaned BGP CRDs and VRFs

privateip:
Add a periodic garbage collection pass to galactic-router that cleans up stale BGPAdvertisement and BGPVRFInstance CRDs, plus orphaned VRF interfaces left behind when containers are force-terminated and CNI DEL never fires.

The GC runs every 5 minutes by default (configurable via flags or env vars). It operates in two phases: first it removes BGP CRDs whose associated container network namespace no longer exists on the node, then it removes VRF interfaces whose corresponding BGPAdvertisement CRD is missing.

- Detect orphaned BGPAdvertisement CRDs by checking if the container netns bind-mount still exists under /var/run/netns
- Remove corresponding BGPVRFInstance CRDs that share the same name
- Detect orphaned VRF interfaces by matching Galactic naming pattern and checking for a live BGPAdvertisement CRD
- Export vrf.ListVRFLinks() from the VRF plumbing package
- Add gc-namespace and gc-interval CLI flags with env var bindings
- Run an initial GC pass immediately on startup, then on each tick

### DIFF
--- a/cmd/galactic-router/root.go
+++ b/cmd/galactic-router/root.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"net"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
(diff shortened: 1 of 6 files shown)


scotwells:
Any reason we can't use standard k8s garbage collection for this?

privateip:
Part of the gc operation is to clean up orphaned host entries (e.g. vrfs) which means the gc process has to run on the node itself.

scotwells (approved):
@privateip makes sense. Did you ever play with the concept of exposing those as CRDs?

privateip:
> Did you ever play with the concept of exposing those as CRDs?

Yes and i need to get back to it as its not quite done yet

PR #208, opened 2026-07-08
feat: deploy galactic-cni via a hostPath installer daemonset

privateip:
galactic-cni previously had no real deployment path — it was compiled directly into the kindest-node-galactic Kind image and only worked in containerlab, with a script that manually minted a long-lived ServiceAccount token to build its kubeconfig. Add a production DaemonSet, modeled on Calico/Cilium's CNI installers, and rewire containerlab to use it instead of baking the binary into the node.

- Add deploy/galactic-cni/{daemonset,configmap,rbac,serviceaccount}.yaml: an initContainer copies galactic-cni and host-device onto the host via hostPath and writes a wrapper that resolves the node name from the downward API, then a lightweight main container refreshes the kubeconfig from the pod's own ServiceAccount token on an interval
- Bracket IPv6 literals when building the kubeconfig server URL, since KUBERNETES_SERVICE_HOST is an IPv6 address on IPv6-only clusters
- Mirror the manifests into deploy/containerlab/resources/cni with imagePullPolicy: Never and a nodeAffinity excluding control-plane nodes, and rewrite deploy-cni.sh to apply them instead of manually pushing a kubeconfig
- Drop the golang builder stage and CNI install block from the Kind node image and install.sh now that the DaemonSet handles it
- Wire build:galactic-cni and its image load/cleanup into the containerlab Taskfile, and update README/ARCHITECTURE docs

### DIFF
--- a/deploy/containerlab/resources/cni/daemonset.yaml
+++ b/deploy/containerlab/resources/cni/daemonset.yaml
@@ -1,0 +1,92 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: galactic-cni
+  namespace: galactic-system
+  labels:
+    app.kubernetes.io/name: galactic-cni
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: galactic-cni
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: galactic-cni
+    spec:
+      serviceAccountName: galactic-cni
+      tolerations:
+        - operator: Exists
+      # Only worker nodes run VPC pods; skip the Kind control-plane nodes so
+      # the galactic-cni:latest image only needs loading onto workers.
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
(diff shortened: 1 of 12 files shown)


ecv (approved):

deploy/containerlab/resources/cni/daemonset.yaml
scotwells, line 1: Why duplicate these resources twice instead of just referencing to deployment manifests at `deploy/galactic-cni`?

privateip: thats a good point.  will do another pr to address that

scotwells: @privateip could we also keep deployment manifests at `config` in the root of the directory? What's pretty standard across our services and typically what's found with kustomize based service deployments. Open to changing the standard to `deploy/` long term if that's a better fit.

privateip: I will move the manifests to `config` to keep them consistent.

privateip: fixed in https://github.com/datum-cloud/galactic/pull/210

PR #217, opened 2026-07-09
fix(gc): stop deleting live BGP CRDs/VRFs out from under running pods

privateip:
## Summary

Three stacked root causes, all contributing to VPC pods losing cross-site connectivity ~5 minutes after every deploy (the GC controller's default interval), when it wrongly deleted live `BGPAdvertisement`/`BGPVRFInstance` CRDs and kernel VRFs as "orphaned":

1. **Wrong liveness check** — GC reconstructed a netns bind-mount name from the container ID and looked for it under `/var/run/netns`. Real netns mounts are named by the container runtime's own convention (e.g. containerd's `cni-<uuid>`), unrelated to container ID, so the check always returned false. The CNI plugin now records the exact netns path it was invoked with as a new annotation at attach time; GC checks that exact path instead of guessing, and only considers a CRD orphaned once *none* of its (possibly several, from pod churn) recorded paths are alive.
2. **No filesystem visibility** — the router DaemonSet had no `hostPath` mount for `/var/run/netns` at all, so even a correct check had nothing to look at. Added the missing volume mount.
3. **No node scoping** — `BGPAdvertisement`/`BGPVRFInstance` CRDs are namespace-scoped, not node-scoped, but GC's liveness check only has visibility into its own node's kernel/filesystem state. On iad — the only site running two router pods (tenant + tenant-control) against the same namespace — each was judging the *other's* live resources against its own local state, always finding them "orphaned" and deleting them. Single-router sites (dfw, sjc) never hit this. `gc.RunGC` now filters every CRD to only those owned by a `BGPRouter` targeting the current node before judging orphan status.

Also removes the redundant static per-site `BGPAdvertisement` manifests (`resources/bgp/tenant/$site/bgpadvertisement-vpc{10,20}.yaml`) discovered stale during this investigation — the CNI's dynamic per-pod advertisement already sets the same SID as its EVPN `GWIPAddress` on every attach.

**Depends on `refactor/config-kustomize`** — base is set to that branch, not `main`, since this builds on its `config/router/base/daemonset.yaml`. GitHub will offer to retarget to `main` once that PR merges.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `gofmt -l` clean.
- [x] `go test ./internal/gc/... ./internal/controller/...` passes.
- [x] Live-verified against a running containerlab deployment: rebuilt and redeployed `galactic-router`/`galactic-cni`, restarted VPC workloads, and confirmed all three sites (including iad's two-router topology) held stable — zero restarts, zero GC/panic activity — across multiple full GC intervals (5+ minutes each), with working cross-site pings and intact VPC isolation.


ecv:
READY TO MERGE!? this repo needs branch protections and the rest of the works

scotwells (approved):

PR #224, opened 2026-07-10
Publish per-binary galactic-cni and galactic-router images

privateip:
## Summary
- Add `.github/workflows/publish.yaml`, modeled on the `compute` repo's `publish.yaml`: publishes `galactic-cni` and `galactic-router` as separate images via the reusable `datum-cloud/actions` docker-publish workflow, plus `config/` as a Kustomize OCI bundle, on every push and published release.
- Add `containers/galactic-router/Dockerfile` (distroless-static, mirrors `deploy/containerlab/containers/galactic-router/Dockerfile`) so `galactic-router` actually gets built and published.
- Repoint `config/cni/daemonset.yaml` and `config/router/base/daemonset.yaml` from the old shared `ghcr.io/datum-cloud/galactic:latest` to `ghcr.io/datum-cloud/galactic-cni:latest` / `ghcr.io/datum-cloud/galactic-router:latest`.
- Update `docs/agents/ARCHITECTURE.md` and `README.md` to drop the now-stale "no production image build path" constraint.

This fixes the gap recorded in `ARCHITECTURE.md`'s Known Constraints: the previous shared image only ever built `galactic-cni` while the router DaemonSet ran `command: [/galactic-router]` against it — the image advertised a binary it never built. Splitting into two per-binary images (matching what `deploy/containerlab/` already does for local dev) fixes this directly instead of resurrecting the broken shared-image pattern.

## Test plan
- [x] `task lint` passes
- [x] `task build` passes
- [x] `docker build -f containers/galactic-router/Dockerfile .` succeeds; ran the resulting binary and confirmed `--help`/flag output
- [x] `kubectl kustomize config/` renders the four container specs with the correct new image references
- [ ] CI run on this PR (lint/build/test-unit/test-e2e via `ci.yaml`)
- [ ] `publish.yaml` will only actually push images once merged/run on `push`/`release` — not exercised by this PR itself


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,42 @@
...
+  publish-kustomize-bundles:
+    permissions:
+      id-token: write
+      contents: read
+      packages: write
+    uses: datum-cloud/actions/.github/workflows/publish-kustomize-bundle.yaml@v1.20.0
+    with:
+      bundle-name: ghcr.io/datum-cloud/galactic-kustomize
+      bundle-path: config
+    secrets: inherit
```
scotwells, line 48: You'll want to leverage this new feature I pushed to the shared action that lets you stamp the kustomize bundle with multiple image versions. This will embed the version tag in the kustomize bundle.

See https://github.com/datum-cloud/actions/pull/89

scotwells (approved):

PR #251, opened 2026-07-20
cni: annotate NAD with host interface name

privateip:
Annotate the NetworkAttachmentDefinition with the deterministic host-side interface name under \`k8s.v1.cni.cncf.io/host-interface\`. This allows downstream processes to discover the interface name before pod launch without querying VPCAttachment CRs.

**Changes:**
- New \`internal/cni/nad.go\` with \`annotateNAD()\` and \`parsePodNamespace()\`
- Unit tests for \`parsePodNamespace\` covering edge cases
- RBAC rule for patching \`network-attachment-definitions\`
- Best-effort integration in \`cmdAdd\` — tolerates NotFound and Conflict
- Deduplicate k8s client init by storing on \`resourceTracker\`

**Why:** Interface name is deterministic from VPC+VPCAttachment but downstream processes need a queryable resource. Annotating the NAD is cleaner than creating a dedicated CRD for this single identifier.

Generated with Qwen Code (1-shotted by Qwen-Coder)

### DIFF
--- a/internal/cni/nad.go
+++ b/internal/cni/nad.go
@@ -43,10 +43,14 @@ func parsePodNamespace(cniArgs string) string {
 }
 
 // annotateNAD patches the NetworkAttachmentDefinition with the host interface
-// name. This is a best-effort operation — failure is logged but not returned.
-func annotateNAD(ctx context.Context, k8s client.Client, nadName, nadNamespace, hostInterface string) {
+// name. The NAD is expected to already exist (created by the external VPC
+// operator before the CNI is invoked), so a not-found response is a hard
+// failure rather than something to tolerate. A conflict response is the one
+// case treated as non-fatal: it means the annotation was already applied by a
+// previous invocation.
+func annotateNAD(ctx context.Context, k8s client.Client, nadName, nadNamespace, hostInterface string) error {
 	if nadNamespace == "" {
-		return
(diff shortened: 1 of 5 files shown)


internal/cni/nad.go
```diff
@@ -0,0 +1,72 @@
...
+	err := k8s.Patch(ctx, nad, client.RawPatch(types.JSONPatchType, []byte(patch)))
+	if err != nil {
+		// Tolerate not-found (NAD managed by external operator that may not
+		// have created it yet) and conflict (annotation already exists from
+		// a previous invocation).
+		if !apierrors.IsNotFound(err) && !apierrors.IsConflict(err) {
+			slog.Warn("annotate NAD: patch failed",
+				"name", nadName, "namespace", nadNamespace, "err", err)
+		}
+		return
```
scotwells: Why wouldn't this hard fail?

privateip: Right now its possible that the NAD isn't there and it probably should fail hard.   I'm working on a new PR that will automate NAD creation so technically this should never happen.    That said, its probably good practice to force this scenario to hard fail.

privateip: fixed in 8ef8fa3

scotwells (approved):

PR #263, opened 2026-07-27
feat(vmtap-cni): add chained CNI plugin for kraftlet/Cilium tap redirect

privateip:
## Summary

- Adds `vmtap-cni`, a standalone CNI plugin chained after Cilium's own CNI plugin, giving a Unikraft microVM (managed by `kraftlet`) access to the pod's real Cilium-assigned identity via a tap device and bidirectional `tc-mirred` redirect — the same mechanism Kata's `tcfilter` endpoint type uses. Unrelated to `galactic-cni`'s VPC/SRv6 dataplane: no VRF/BGP involvement, no shared config surface.
- Tap MTU is set to the pod's resolved *route* MTU (read from the kernel routing table), not `eth0`'s link MTU, addressing the documented Cilium/Kata MTU mismatch. `eth0` itself is never modified.
- Adds a `config/vmtap` DaemonSet with a `patch-conflist` loop (`vmtap-cni patch-conflist`) that appends the plugin into Cilium's conflist and keeps re-checking it, since DaemonSet startup ordering across Cilium and vmtap-cni isn't guaranteed.
- Wires a third binary (`bin/vmtap-cni`) into `Taskfile.yaml`'s build task and adds `containers/vmtap-cni/Dockerfile` (e2e-only, mirroring the existing `galactic-cni` Dockerfile pattern — not yet wired into `scripts/ci.sh`, since real end-to-end validation needs a Cilium-enabled cluster).
- Adds `docs/vmtap-cni/configuration.md` covering config fields, MTU handling, the kraftlet hand-off convention, and the still-unvalidated Cilium caveats (tc/bpf hook ordering, socketLB, Envoy sidecar interaction).

### Explicitly out of scope / open items (see plan section 7 and the doc's "Open items" section)

- Verified tc/bpf filter priority against a real Cilium version/datapath mode (veth+tc vs. netkit vs. XDP native routing).
- The pod-level trigger signal (RuntimeClass vs. annotation) deciding which pods get this conflist entry — the DaemonSet currently gates on a placeholder node label.
- Confirmation of the `sandbox: CNI_CONTAINERID` convention against kraftlet's actual CRI/containerd integration.
- Cluster-wide `socketLB.hostNamespaceOnly=true` Cilium config (owned by whoever manages the Cilium install, not this plugin).

## Test plan

- [x] `task build` — all three binaries (`galactic-cni`, `galactic-router`, `vmtap-cni`) build
- [x] `go vet ./...`
- [x] `task lint` (golangci-lint + yamlfmt) — 0 issues
- [x] `bash scripts/ci.sh unittest` — full suite passes; new root-gated netns/tap/tc tests skip cleanly without CAP_NET_ADMIN (same convention as `internal/cni`)
- [x] `kubectl kustomize config/` and `kubectl kustomize config/vmtap` build cleanly
- [ ] End-to-end validation against a real Cilium-managed cluster (NetworkPolicy, Service routing, Hubble attribution, MTU, throughput) — deferred, tracked as an open item

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -99,6 +99,7 @@ tasks:
     cmds:
       - go build -ldflags "{{.LDFLAGS}}" -o bin/galactic-cni ./cmd/galactic-cni
       - go build -ldflags "{{.LDFLAGS}}" -o bin/galactic-router ./cmd/galactic-router
+      - go build -ldflags "{{.LDFLAGS}}" -o bin/vmtap-cni ./cmd/vmtap-cni
       - GOBIN={{.LOCALBIN}} go install github.com/containernetworking/plugins/plugins/main/host-device@v1.9.1
 
   ##
--- a/cmd/vmtap-cni/main.go
+++ b/cmd/vmtap-cni/main.go
@@ -1,0 +1,108 @@
+// Copyright 2026 Datum Cloud, Inc.
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/containernetworking/cni/pkg/version"
+	"github.com/spf13/cobra"
+	"golang.org/x/term"
+
+	"go.datum.net/galactic/internal/metadata"
+	"go.datum.net/galactic/internal/vmtap"
+)
+
+const (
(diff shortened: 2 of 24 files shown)


scotwells (approved):
Wonder if this should live in the unikraft-provider repo long term so all unikraft specific stuff is in one place?

privateip:
> Wonder if this should live in the unikraft-provider repo long term so all unikraft specific stuff is in one place?

Yeah probably

PR #268, opened 2026-07-30
feat(webhook,cni,gc): automate VPC attachment provisioning

privateip:
## Summary

Pods requesting a VPC via the `galactic.datumapis.com/vpc` annotation previously required a `NetworkAttachmentDefinition` (NAD) to be created manually ahead of time, and there was no automated path from "pod wants a VPC" to a working attachment.

A new `galactic-webhook` mutating admission webhook now handles the pod-facing side: on Pod CREATE it allocates a free VPCAttachment ID for the named VPC and creates the NAD itself, then patches the pod to attach via it. `galactic-cni` handles the data-plane side: it now creates the `VPCAttachment` CR (Spec and Status) at the same point it already creates `BGPVRFInstance`/`BGPAdvertisement`, once real IPAM addresses are known — so the CRD's required fields are always populated from actual allocation data, not guessed at admission time.

The GC controller gained matching cleanup passes: orphaned NADs are reclaimed by reference-counting against live pods' networks annotations, and orphaned VPCAttachments are reclaimed via their own `Status.PodName`.

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -99,6 +99,7 @@ tasks:
     cmds:
       - go build -ldflags "{{.LDFLAGS}}" -o bin/galactic-cni ./cmd/galactic-cni
       - go build -ldflags "{{.LDFLAGS}}" -o bin/galactic-router ./cmd/galactic-router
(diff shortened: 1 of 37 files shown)


privateip:
Closing this PR after discussions with @scotwells.  We are working on a different way to handle VPC definition and attachment of workload instances.

PR #270, opened 2026-07-31
fix: Make IPAM default route idempotent on ADD retries

privateip:
## Summary

galactic-cni's IPAM step adds a default route toward the attachment's gateway on every CNI ADD. If a caller retries ADD against the same netns without an intervening DEL — which happens in practice when an earlier ADD attempt is aborted before its own rollback runs — the route already exists and ADD fails with `configure IPAM: add default route via <gateway>: file exists`, leaving the pod stuck in `ContainerCreating` on every subsequent retry.

The IPAM route/address install now checks existing kernel state first: if the address or default route already match what's being installed, the call is a no-op success. If a default route exists via a *different* gateway, that's a real misconfiguration and now fails loudly instead of being silently papered over or overwritten.

Also hardened the DEL path: the error from delegating DEL to the host-device plugin (the mechanism that actually flushes the route/address, by moving the guest interface back out of the netns) was previously discarded entirely. It's now logged, so a failed cleanup — the underlying cause of the stale route in the first place — is visible instead of invisible.

### DIFF
--- a/internal/cni/netns.go
+++ b/internal/cni/netns.go
@@ -62,7 +62,11 @@ func configureInterfaceInNetns(
 
 // addAddrAndDefaultRoute assigns ipNet to link and, if gateway is set,
 // installs a default route via it. No-op when ipNet is nil (family not in
-// use for this attachment).
+// use for this attachment). Both the address and the route are added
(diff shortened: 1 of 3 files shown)


drewr (approved):
Let's try it

PR #273, opened 2026-08-02
feat(srv6): cut over to a TC-BPF uFMT 48+16 uSID datapath

privateip:
## Summary

- Replaces galactic's per-endpoint `seg6local` uSID ingress route model (one static kernel route per `(VPC, VPCAttachment)`, O(N) FIB entries per node) with a new eBPF/TC-BPF datapath (`internal/plumbing/ebpf/`) implementing the `uFMT 48+16` uSID carrier from datum-cloud/enhancements#740 (RFC 9800 REPLACE-CSID / Option 2) — O(1) FIB cost per node regardless of tenant count.
- `ComputeSID` (`internal/plumbing/srv6/usid.go`) now emits the real uFMT 48+16 layout, backed by a new local per-node Argument allocator (`allocateArgument`, `internal/cni/bgp.go`) sourced from `BGPVRFInstance` CRD state.
- `galactic-cni` registers/unregisters each attachment's `vrf_table` entry on ADD/rollback; a new GC sweep (`internal/gc`) reclaims entries orphaned by deleted `BGPVRFInstance`s.
- Direct cutover, not a phased rollout: not in production yet, so the legacy `seg6local` path and its feature flags (`GALACTIC_CNI_ENABLE_EBPF_DATAPATH`, `GALACTIC_CNI_EBPF_OBSERVE_ONLY`) are deleted outright rather than kept behind a flag/observe-only mode.
- Fixes two correctness bugs found deploying this to a live ContainerLab fabric: the netlink interface watcher never re-verified an interface an external event (an FRR restart) had silently detached from, and `RouteEgressAdd`'s full SEG6 encap mode pushed a Segment Routing Header that `usid.c`'s fixed-width header strip didn't account for. Also adds a `bpf_fib_lookup` neighbor-priming fix for pods whose address never otherwise triggered ARP/NDP.
- Carries the SRv6 SID for IPv4-family EVPN Type 5 routes via a new RFC 9252 Prefix-SID path attribute (`internal/runtime/gobgp`), since `GWIPAddress` can only carry a gateway matching its own NLRI's address family and was silently dropping the SID for IPv4 prefixes.
- Corrects stale docs/comments claiming `vpc` is a 48-bit identifier (it's 16-bit, cluster-scoped) and adds a package-level `doc.go` overview for `internal/plumbing/ebpf/`.


docs/ebpf-datapath-sequence.md
scotwells, line 1: These diagrams aren't loading for me

scotwells (commented):
What kind of telemetry do we get out of the eBPF program? Could we document any of the operational telemetry as documentation in the repo?

ecv:
I am letting Claude post for me on grounds _woof_. 

=====

Read the whole thing. The design holds up and the C is careful — three of these commits are live-fabric fixes, which shows. But this becomes the only forwarding path, and the packet path runs nowhere in CI.

1. The health check never reaches kubelet. `Run` publishes datapath status under the `ebpf-datapath` gRPC service; the DaemonSet's probes omit `service:`, so kubelet polls `""`, which `Run` sets SERVING at startup and never touches again. The FRR-restart filter detach this PR fixes still won't fail a probe or restart the pod. Add `service: ebpf-datapath` to both probes.

2. `usid.c` never reads `ip6->nexthdr`. It assumes byte 40 begins the inner packet, so any extension header — an SRH from a node still on full encap, a fragment header — lands in the version nibble and drops as `UNKNOWN_INNER_VERSION`. That is the bug the ENCAP_RED commit chased. Check nexthdr for 4 or 41 and count it apart.

3. Unit Tests logs 51 `requires root` skips; every `BPF_PROG_TEST_RUN` case in `usid_test.go` is one of them. E2E pins the maps and runs CNI ADD, but never forwards a packet. So the 623 lines of datapath ran only on your ContainerLab fabric. Runners have sudo — a second job running the suite as root closes this.

4. `BPF_FIB_LKUP_RET_FRAG_NEEDED` cannot fire. The kernel MTU-checks only when `params->tot_len` is set, and `fib_params.tot_len` stays zero. Set it from the inner header, or drop the counter.

5. The committed `usid_bpfel.o`/`usid_bpfeb.o` can drift from `usid.c` unseen. The image rebuilds them, so what ships is right, but `go test` loads the blob. Gitignore them, or regenerate in CI and `git diff --exit-code`.

6. Twenty-eight comments cite `.local/plan-ebpf-xdp-usid-datapath.md` and its milestone file. Neither is in the repo. They carry the reasoning — R2, R8, §4.2 — that a reviewer needs to judge the code. Move it into `docs/` or the enhancement PR.

7. `daemonset.yaml` still says the datapath stays inert until `GALACTIC_CNI_ENABLE_EBPF_DATAPATH=true`. This PR deletes that flag. The line sits directly above the privilege grant it contradicts.

8. Any packet to `Block:NodeID::/64` carrying an unregistered Function dies as `TC_ACT_SHOT` on the underlay link. Right for a dedicated SID /64, silent for whoever numbers a node address inside one. Worth a line in the docs.

9. Smaller: `registerEBPFDatapath` narrows `bgp.nodeID` to uint16 unchecked, while `ComputeSID` rejects the same value — a bad NodeID registers a locator no SID will ever match. And preflight probes the fib-lookup TBID parameter but not seg6 reduced encap, whose absence drops every cross-region packet.

The Argument allocation reads right: both racers failing and retrying beats a tie-break, and per-node scoping holds because the RD carries routerID and the RT carries the VPC.

privateip:
> What kind of telemetry do we get out of the eBPF program? Could we document any of the operational telemetry as documentation in the repo?

I just spoke with @0xmc about this yesterday and I have an action item to document all of the currently available touch points for observability.  I will tackle that later this week

docs/ebpf-datapath-sequence.md
privateip: fixed

privateip:
Split this into more digestible PRs

PR #282, opened 2026-08-05
feat: eBPF codegen toolchain for the uSID TC-BPF program

privateip:
## Summary

Second PR in the eBPF uSID datapath stack (on top of #281, which adds the `usid.c` program this generates from).

Teaches `task build` to regenerate the TC-BPF program's Go bindings via `go generate` (a new `build:ebpf` step ahead of `build:binaries`) and the CNI container image to do the same at image-build time — `clang`/`llvm` stays a build-time-only dependency and never ships in the runtime image.

Also adds a root-privileged rerun of the unit test suite in CI (`test-unit-root`): the existing `test-unit` job runs as an unprivileged user, so every test gated on `requireRoot(t)` — including the `BPF_PROG_TEST_RUN` cases in `usid_test.go` — skips instead of running. And a step that diffs the regenerated eBPF artifacts against the committed ones, so `usid.c` and its compiled output can't silently drift apart.

Note: `test-e2e`'s dependency list isn't updated to include `test-unit-root` yet — that lands in the deploy/e2e PR later in this stack, once there's an eBPF datapath for e2e to actually depend on.

## Test plan

- [x] `go build ./...`
- [x] `task --list` (Taskfile parses, `build:ebpf` task registered)
- [ ] CI itself exercises the new jobs once opened (no `clang` available in this local sandbox to run `build:ebpf` directly)

Part of the eBPF uSID datapath cutover stack (base: #281).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,6 +53,47 @@ jobs:
           files: coverage.out
           fail_ci_if_error: false
 
+  # test-unit above runs as the unprivileged runner user, so every test
+  # gated on requireRoot(t) (BPF_PROG_TEST_RUN cases in
+  # internal/plumbing/ebpf/prog/usid_test.go, plus the real-kernel BPF map
+  # load/attach/GC-sweep tests across internal/plumbing/ebpf/attach,
+  # internal/plumbing/ebpf/usidmap, internal/gc, internal/cni, and
+  # internal/installer) skips instead of running -- the only unit-level
+  # exercise of usid.c's 623 lines is otherwise the containerlab fabric.
+  # GitHub-hosted runners grant passwordless sudo, so re-running the exact
+  # same suite as root here closes that gap by actually loading the BPF
+  # program into the runner's kernel and driving it through
+  # BPF_PROG_TEST_RUN instead of skipping.
+  test-unit-root:
+    name: Unit Tests (root)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: actions/setup-go@v7
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - uses: arduino/setup-task@v3.0.0
+        with:
+          version: 3.x
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Mark workspace as safe for root
+        # actions/checkout leaves the repo owned by the runner user; git
+        # refuses "dubious ownership" once root reads it below, which
+        # would otherwise make `go test`'s VCS stamping fail.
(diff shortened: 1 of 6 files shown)


ecv:
Same disclosure as on #281 — ecv reviewed this through an AI shell, apologies. Questions, not verdicts.

1. The drift check pins correctness to whatever clang `ubuntu-latest` happens to ship. `apt-get install clang` is unversioned, and BPF object bytes move with the compiler (BTF, debug info, section layout), so the next runner-image clang bump makes `git diff --exit-code` fail on every PR — including ones nowhere near `usid.c` — with a diff nobody introduced. Is it worth pinning an explicit `clang-NN` here, or comparing something less brittle than raw bytes?

2. The Dockerfile regenerates the object during the image build, so the `.o` that ships is the builder image's clang output, not the committed blob that CI verified and reviewers read. That leaves three artifacts that can differ: committed, CI-regenerated, shipped. What guarantees the shipped datapath is the one that was reviewed? If the answer is "nothing, and that's fine because it's regenerated from the same source," is committing the blob still earning its keep — and if it is, would embedding the committed object in the image build (rather than regenerating) be the stronger guarantee?

3. `build:binaries` going `internal: true` makes clang a hard requirement for `task build`, so a contributor touching unrelated Go now needs the eBPF toolchain or gets a hard failure. Is that intended? A `build:ebpf` that warns and skips when the committed artifacts are already current would keep the toolchain optional for everyone except the people actually editing `usid.c`.

4. The Dockerfile installs `linux-libc-dev` alongside clang/llvm; the CI step installs only `clang llvm`. `doc.go`'s `-idirafter` workaround exists specifically to find that package's headers, so CI seems to be relying on it being preinstalled on the runner image. Should the two dependency lists match, so a slimmer runner image doesn't break the build in a way the Dockerfile already anticipated?

5. `test-unit-root` isn't in anyone's `needs` (`test-e2e` is `[lint, test-unit, build]`) — is it in the required-checks list for branch protection? If not, the job can go red without blocking a merge, which would make the coverage it adds advisory only. Related to the note you already left about wiring it into `test-e2e` later.

6. It re-runs the entire suite as root rather than just the root-gated packages. Beyond the duplicated runtime, any test that asserts a permission-denied path behaves differently as root. Worth scoping to the eBPF packages, or is whole-suite-as-root deliberate?

One note rather than a question: because this PR targets `pr0-ebpf-prog` and the workflow triggers on `pull_request: branches: [main]`, CI hasn't run here at all — only the CLA and image-publish jobs. So the clang install, the drift check, and `test-unit-root` itself are all still unexercised, which matches your unchecked third box. Same applies to every PR in the stack until the base lands on `main`.

ecv:
this one's got failing tests @privateip

ecv (approved):
can you tech-debt ticket yourself to move the comments into documentation, AFTER merge?

PR #284, opened 2026-08-05
fix: Merge BGPAdvertisement prefixes across shared containers

privateip:
## Summary

A single BGPAdvertisement CRD is keyed by `(vpc, vpcAttachment)` alone, so multiple containers attaching under the same VPCAttachment on a node — a second pod, or a second interface reusing it — share one CRD. `publishBGPStateK8s` was building `Spec.Prefixes` from only the container currently being processed, so a second container's CNI ADD would silently drop the first still-live container's already-published prefix from the spec.

`allAdvertisedPrefixes` fixes this by recomputing `Spec.Prefixes` from every subnet annotation currently present on the CRD (each container's durable per-container record) instead of just the one being processed, so every live container's prefix stays present regardless of ADD order.

Independent of the eBPF uSID datapath work landing in a separate stack — no shared imports or call-site overlap beyond this file's existing `publishBGPStateK8s`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./internal/cni/...`
- [x] `go test ./internal/cni/...` (new `TestAllAdvertisedPrefixes*` cases, including the multi-container regression case)
- [x] `task lint` (0 issues)

### DIFF
--- a/internal/cni/bgp.go
+++ b/internal/cni/bgp.go
@@ -10,7 +10,9 @@ import (
 	"fmt"
 	"log/slog"
 	"net"
+	"sort"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ecv (approved):
seems legit. we should do a pass pulling all the comments out into design docs later

PR #288, opened 2026-08-05
feat: Grant CNI DaemonSet BPF privileges and wire up e2e

privateip:
## Summary

Final PR in the eBPF uSID datapath stack — makes the feature actually deployable and end-to-end testable.

Grants the CNI DaemonSet's `credential-refresh` container `CAP_BPF`/`CAP_NET_ADMIN` plus a `bpf-fs` hostPath mount (`/sys/fs/bpf`), and adds an `ebpf-datapath` gRPC health sub-service to its liveness/readiness probes — required unconditionally now that this container also hosts the eBPF/TC-BPF uSID datapath's control daemon (#287); there's no flag to gate this behind since the datapath is the only forwarding path.

Pins `GALACTIC_CNI_EBPF_INTERFACES=eth1` for the containerlab topology, where every lab node is dual-homed and interface auto-detection would otherwise pick the wrong (management-bridge) link over the real transit-fabric one.

Mounts bpffs on Kind nodes in CI and updates `TestCNITapInterface` to run a privileged e2e pod with its own `bpf-fs` mount, starting the eBPF control daemon and waiting for `vrf_table` to be pinned before exercising CNI ADD — the datapath being the only forwarding path means the e2e pod needs the same maps a production node's DaemonSet would already have pinned. Adds `test-unit-root` to `test-e2e`'s CI dependency list.

> [!NOTE]
> This is stacked on both #286 (CNI eBPF cutover + gateway fix) and #287 (GC/installer wiring) via a merge commit, so the diff GitHub shows here temporarily includes both of those until they merge upstream. Review should focus on the deploy/e2e files: `config/cni/daemonset.yaml`, `deploy/containerlab/resources/cni/daemonset-patch.yaml`, `scripts/ci.sh`, `tests/e2e/e2e_test.go`, and the `test-e2e` dependency-list line in `.github/workflows/ci.yaml`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...` (including `tests/e2e`, compiles clean)
- [x] `task lint` (0 issues)
- [ ] `task test:e2e` needs a live Kind cluster lifecycle — not run in this sandbox; CI will exercise it

Part of the eBPF uSID datapath cutover stack (base: #286, #287).


ecv (approved):
let's do a pass through this to remove comments and xfer them into docs, AFTER it merges?

PR #298, opened 2026-08-07
fix: Use reduced SRv6 encap so eBPF ingress stops dropping packets

### LINKED ISSUE #297: Cross-cluster pod traffic silently dropped by eBPF uSID ingress datapath
## What needs to happen

Pods in different clusters/sites can't reach each other over the SRv6 data plane. Every packet that reaches the destination node is silently dropped — no ICMP, no error, nothing in the CNI or router logs.

## Why this matters

This breaks all inter-cluster VPC connectivity for any VPCAttachment spanning more than one node's SRv6 domain, which is the core use case the SRv6 data plane exists for.

## Desired outcome

A pod on one node can reach a pod on another node/site over the SRv6-encapsulated path, with the same reliability as before the eBPF uSID datapath (PR #283) replaced the kernel `seg6local` ingress route model.

## Notes

Root-caused in a containerlab reproduction: the egress route installer builds a non-reduced SRv6 encap route, which the kernel always wraps in a Segment Routing Header — but the eBPF ingress program introduced in #283 requires no SRH at all for the single-segment case, and drops anything else (`DROP_REASON_UNEXPECTED_NEXTHDR`) with `vrf_table`'s own hit counters confirming every claimed packet was dropped.

privateip:
## Summary

Pods in different clusters couldn't reach each other over the SRv6 data plane — every packet arrived at the destination node and was silently dropped, with no error visible anywhere in the control plane.

The egress route installer was building its SRv6 route with a non-reduced encap mode, which the kernel always wraps in a full Segment Routing Header. The eBPF ingress program that replaced the old kernel `seg6local` route model in #283 requires no such header for the single-segment case this installer always produces, and drops anything else. This switches the egress route to reduced encap, which the kernel renders as a bare encapsulation with no extra header — matching what the ingress side expects. Confirmed on the wire, for both IPv4 and IPv6 inner payloads, that this produces exactly the format the ingress side requires, and that cross-cluster ping now succeeds end-to-end.

## Test plan

- [x] `go build ./...` and full unit test suite pass
- [x] Reproduced the failure and confirmed the fix live in a 3-cluster containerlab lab (dfw/iad/sjc): rebuilt and redeployed `galactic-router` with the fix, all 6 cross-cluster pod-to-pod ping pairs went from 100% to 0% packet loss
- [x] Confirmed via kernel packet capture, before and after, that the new route's outer header carries no Routing Header for both IPv4 and IPv6 inner payloads

Fixes #297


scotwells (dismissed):
What kind of observability do we have around galactic where we could detect this? Assume we have metrics around packets dropped?

scotwells (approved):

PR #300, opened 2026-08-07
fix(srv6): use Gw instead of Via for same-family egress routes

privateip:
RouteEgressAdd unconditionally attached the resolved next-hop via
netlink.Route.Via, which the kernel only accepts when Via address family
differs from Dst's. IPv4 VPC prefixes egressing over the IPv6
SRv6 underlay need exactly that cross-family Via — and already worked —
but IPv6 VPC prefixes share Dst's family and must use the plain Gw
field, or the kernel rejects the route with EINVAL.

Pick Via for cross-family (IPv4 prefix) and Gw for same-family (IPv6
prefix). Add a root test that installs both prefix families against a
real kernel routing table and verifies the installed attribute matches.


ecv (approved):
comments hrrnnnngh

privateip:
Ignore the build error, I have a PR coming that is going to unwind all of that and get the compiled code out of the repo.

PR #302, opened 2026-08-07
build(ebpf): stop committing bpf2go-compiled artifacts

privateip:
## Summary

`internal/plumbing/ebpf/prog`'s `usid_bpfel.o`/`usid_bpfeb.o` (and their bpf2go-generated `.go` bindings) were committed to git and `go:embed`'d into the binaries. Unlike other generated code (`*.pb.go`), these embed a compiled binary blob, so a stale commit could silently drift out of sync with `usid.c` with nothing but a byte-diff to catch it.

This gitignores the four generated files and requires every build site to regenerate them from `usid.c` via bpf2go instead of shipping a checked-in copy.

## Changes

- **`Taskfile.yaml`**: `build:ebpf` now unconditionally regenerates (clang required, via a new `require-clang` check) instead of falling back to a committed copy when clang is missing. Reordered `build` so `build:ebpf` runs before `fmt`/`vet`, which need the embed target to compile. Added `build:ebpf` as a dependency of `lint`, `lint-fix`, `test:unit`, `test:unit-root`, and `test:e2e`.
- **`ci.yaml`**: installs the pinned `clang-18`/`llvm-18`/`linux-libc-dev` toolchain in the `lint`, `test-unit`, `test-unit-root`, and `test-e2e` jobs (previously only the `build` job had it). Removed the `build` job's `git diff --exit-code` drift check — there's nothing committed to diff against anymore.
- **`containers/galactic-cni/Dockerfile`** and **`containers/galactic-router/Dockerfile`** (the latter transitively imports `internal/plumbing/ebpf/prog` via `usidmap`): install a clang toolchain and run `go generate` in the builder stage instead of relying on a committed `.o`.
- **`doc.go`**: documents the new convention and why it differs from the repo's committed-generated-code norm.

## Trade-off

clang is now a **hard build requirement everywhere** (dev machines, every CI job, both Dockerfiles) — there's no more build-against-the-stale-committed-copy fallback when it's missing.

## Verification

- `task build`, `task lint`, `task test:unit` all pass from a clean regeneration.
- Both `containers/galactic-cni/Dockerfile` and `containers/galactic-router/Dockerfile` build successfully with the new toolchain step (verified locally via `docker build`).


ecv (dismissed):

0xmc (approved):
Good changes.  I love not have .o files in git.  Some things to consider for a followup:

Make CI DRY-er

Create .github/actions/install-ebpf-deps/action.yaml:

```
  name: Install eBPF build dependencies
  description: Installs the pinned clang/llvm/linux-libc-dev toolchain needed to regenerate internal/plumbing/ebpf/prog's bpf2go output
  runs:
    using: composite
    steps:
      - run: |
          sudo apt-get update
          sudo apt-get install -y clang-18 llvm-18 linux-libc-dev
        shell: bash
```
 
Then each job replaces its 4-line run: block with:

```
        - name: Install eBPF build dependencies
          uses: ./.github/actions/install-ebpf-deps
```

And Claude has recently(?) gotten overly verbose with its comments in code.  My prompt is something along the lines of "no exposition in comments"

PR #303, opened 2026-08-08
refactor(cni): split galactic-cni into veth/tap master plugins

privateip:

- #303 refactor(cni): split galactic-cni into veth/tap master plugins (this PR)

---

## Summary

`galactic-cni` is now veth-only. A new `galactic-tap-cni` binary owns tap workloads (Kata, Firecracker, kraftlet/Unikraft). The `interface_type` config field and its runtime switch are gone from both binaries.

First branch in a stack: `refactor/cni-chain-0-veth-tap-split` → `-1-galactic-ipam` → `-2-galactic-bgp` → `-3-galactic-route` → `-4-installer-docs`. Each branch bases on the previous.

## New shared packages

Extracted so `internal/cnitap` doesn't duplicate or reach into `internal/cni`'s unexported internals:

- **`internal/cni/hostconf`** — `HostConf`/`Load`/node-name API fallback. Match on plugin type is now caller-supplied, not hardcoded to `"galactic-cni"`.
- **`internal/cni/crdnames`** — BGP CRD naming and annotation-key helpers.
- **`internal/cni/nadpatch`** — NAD annotation logic, identical for both interface types.
- **`internal/cniipam`** — IPAM allocation/deallocation. Shared library for now; becomes its own delegated plugin in step 1.
- **`internal/cnibgp`** — BGP/SRv6/eBPF publish logic. Same treatment; becomes its own chain-invoked plugin in step 2.

## Verification

- `task lint` ✅
- `task build` ✅ — builds and stages `galactic-tap-cni` too
- `task test:unit` ✅
- `task test:e2e` — not run. Needs a Kind cluster bring-up (`sudo modprobe vrf`, `kind create cluster`) not attempted here. `installer.Bootstrap`'s static per-node conflist is unchanged, so existing veth-path e2e coverage should hold, but that's unverified — flagging for the reviewer.

Test coverage moved with the code it tests, not dropped: `bgp_test.go`/`bgp_ebpf_test.go` → `internal/cnibgp`, `ipam_ops_test.go` → `internal/cniipam`, `nad_test.go` → `internal/cni/nadpatch`, tap-mode cases from `cni_test.go` → `internal/cnitap`.


privateip:
Thanks for the thorough review, @mattdjenkinson — went through all five points against where this stack has landed since. Two are already handled without further changes needed here:

**`isTapMode()`/`CNI_NETNS_OVERRIDE` dead code** — already deleted, in `725ef87` (this stack's own step 4/#307). Confirmed `cmd/galactic-cni/main.go` no longer has the stdin-peek dance or the `isTapMode()` helper; the comment now in its place explains why it's gone (interface kind is which binary you invoke now, not a config field either binary branches on).

**Missing nil-`tracker.k8s` guard before `cnibgp.PublishBGPState`** — now moot rather than fixed. `e33092c` (step 2/#305, extracting `galactic-bgp`) moved BGP publish entirely out of `internal/cni/ops_add.go` into its own chain-invoked binary. There's now exactly one call site (`internal/cnibgp/ops_add.go`), which constructs its own `k8sClient` immediately before use with an error return on failure — the guard-ordering concern doesn't apply to that shape anymore.

The other three are real and I've landed fixes for them as follow-ups on top of the stack rather than pushing more commits onto this branch, so your review here stays scoped to what you already looked at:

- #315 — fixes the stale `internal/cni/doc.go` package comment (still described the package as veth-or-tap, still listed `tap` as its own subpackage) and promotes `hostconf`/`crdnames`/`nadpatch` from `internal/cni/*` to top-level `internal/*` to match `cnibgp`/`cniipam`/`cnitap`/`cniroute`'s placement — closing the exact "looks like `cnitap` reaches into `cni`'s internals" gap you flagged.
- #316 (stacked on #315) — the `internal/cnitap` vs `internal/cni` duplication in `config.go`/`resource.go`/`ops_check.go`. Extracted the interface-agnostic pieces (parseConf's validation, log setup, k8s client construction, CHECK/STATUS node-level checks, prevResult validation) into a new `internal/cnimaster` package that both master plugins now share, leaving only the genuinely interface-specific bits (which kernel primitive to call, tap's lack of a guest netns to check) in each package. Test coverage moved with the code, including catching that `cnitap`'s own tests had been implicitly relying on `cni`'s tests to cover logic it silently duplicated — exactly the risk you called out.

Both are green on `task lint`/`task build`/`task test:unit`.

mattdjenkinson (dismissed):

mattdjenkinson (approved):

ecv (approved):

ecv (approved):
Approving. Rollback wiring checks out — every `tracker.*` field the old `publishBGPStateK8s` set has a matching `PublishResult` field, and partial results survive a failed retry. Tap cleanup deletes tap, veth cleanup deletes veth.

One thing worth fixing here, two worth knowing about.

**Fix in this PR:** `cmd/galactic-cni/main.go` still has `isTapMode()` reading `interface_type` and setting `CNI_NETNS_OVERRIDE=true`. In a veth-only binary that means a stale tap config skips the same-netns guard and then builds a veth anyway. Deleting the function and its call is the whole fix.

**Sequencing:** `interface_type` is gone from `PluginConf`, so unmarshal now drops it silently — the old binary rejected bad values. Anything the companion operator still emits as `interface_type: tap` + `type: galactic-cni` gets a veth with no error. #307 fixes the docs in this repo; the operator-side switch to `type: galactic-tap-cni` is not covered anywhere in the stack. Worth confirming who lands that and when.

**Behavior change not in the description:** tap CHECK now runs `checkTerminationRoutes`, which the old tap path skipped. Looks like a fix — tap ADD does install those routes — but it can now fail a CHECK that previously could not.

Filed the cleanup items as tech-debt rather than holding the stack: #325, #326.

Also: description says e2e was not run. CI is green on it now, including E2E.

PR #304, opened 2026-08-08
refactor(cni): extract galactic-ipam as a real delegated CNI IPAM plugin

privateip:

- #304 refactor(cni): extract galactic-ipam as a real delegated CNI IPAM plugin (this PR)

---

## Summary

Second branch in the CNI plugin-chain split stack (based on #303). `galactic-ipam` is now a real CNI IPAM delegation target (`github.com/containernetworking/plugins/pkg/ipam.ExecAdd/ExecDel`), execed by `galactic-cni`/`galactic-tap-cni` — not the in-process library call step 0 left it as.

## Explicit IPAM contract

- Whether IPAM runs at all is decided solely by `"ipam"` block presence in a master plugin's own config. No environment variable can trigger or suppress that anymore — the old `GALACTIC_CNI_ENABLE_LOCAL_IPAM` "enforce ipam block" check is gone.
- `ipv6_subnet`/`ipv4_subnet`/`address_families` move inside the `"ipam"` block. Their CIDR validation moved with them, into `galactic-ipam`'s own `parseConf`.
- `ipam.type` now names the delegated binary, not a pool-vs-static mode selector. Mode is decided internally: `static_ip` presence selects static, otherwise `ipv6_subnet`/`ipv4_subnet` select pool mode.
- `enableLocalIPAM` → `GALACTIC_IPAM_ENABLE_LOCAL_IPAM` (`internal/config/ipam.go`), read only by `galactic-ipam` itself, and only fills a default IPv6 pool CIDR when `"ipam"` is present but under-specified — it can't manufacture the block anymore.

## The unlock: IPv6 pool persistence

`PoolAllocator` tracked allocations in memory only — fine for ADD (each pool belongs to exactly one VPCAttachment), useless for DEL (fresh process, nothing to look up). Deallocation used to close that gap by reading the allocated value back out of the BGPAdvertisement CRD annotation `galactic-bgp` wrote — the one thing keeping `galactic-ipam` coupled to `galactic-bgp`'s CRD naming and a Kubernetes client it has no other use for.

Fixed: `PoolAllocator` gets the same on-disk marker file `IPv4PoolAllocator` already had (keyed by containerID, flock-guarded, one shared lock dir for both families — `ipam.DefaultLockDir`). Both allocators gained `DeallocateContainer(containerID)` and a peek-only `LookupContainer(containerID)`. `galactic-ipam` now looks its own allocations up locally — no k8s client, no `crdnames`, no `bgpv1alpha1` import at all.

## Verification

- `task lint` ✅ `task build` ✅ (builds + stages `galactic-ipam` too) `task test:unit` ✅
- `internal/cniipam` at 91% coverage, including a real ADD → CHECK → DEL → CHECK round trip proving the self-contained-by-containerID design holds end to end.
- `task test:e2e` — not run, same caveat as #303.


privateip:
@mattdjenkinson reviewed and addressed your comments

mattdjenkinson (dismissed):

ecv (dismissed):

ecv (dismissed):
Approving — delegation wiring looks right, including rollback's `ExecDel` mid-ADD.

Fast-follow rather than here, so the stack keeps moving: https://github.com/datum-cloud/galactic/issues/327. Old-shape config parses fine, delegates nothing, pod comes up with no addresses and nobody hears about it. Few lines in each master `parseConf`, worth landing before this reaches a real cluster.

Rest filed, not blocking: #328, #329, #330.

CI green including E2E, so the "not run" note is stale.

ecv (approved):

PR #305, opened 2026-08-08
refactor(cni): extract galactic-bgp as its own CNI chain plugin

privateip:

- #305 refactor(cni): extract galactic-bgp as its own CNI chain plugin (this PR)

---

## Summary

Third branch in the CNI plugin-chain split stack (based on #304). `galactic-bgp` is now its own chained CNI plugin, invoked last in the conflist after the master plugin and `galactic-ipam`, instead of in-process BGP/SRv6/eBPF publishing inside `galactic-cni`/`galactic-tap-cni`.

## What moved

`internal/cnibgp` mirrors `cniipam`'s shape: `RunPlugin()`, `PluginConf{VPC, VPCAttachment, Namespace}` parsed from stdin. It reuses `internal/config.CNIConfig`'s `GALACTIC_CNI_*` env vars — shared node-level settings, not a BGP-specific flag.

`galactic-bgp` never touches the container netns. It learns interface kind (veth vs tap) and the allocated IPAM result from `prevResult` alone (`prevresult.go`): two interfaces means veth, one means tap. It publishes BGP state, then passes `prevResult` through unchanged as the last plugin in the chain.

`cmdCheck` is new: verifies the `BGPVRFInstance`/`BGPAdvertisement` CRDs and the eBPF `vrf_table` entry. `cmdStatus` probes the API server, matching STATUS across the chain. `cmdDel` stays a no-op — orphaned VRF/CRD state is reconciled by `galactic-router`'s GC controller, not by any CNI DEL path.

## Rollback scoping

`resourceTracker` in `cnibgp/resource.go` covers only what `galactic-bgp`'s own ADD creates: `BGPVRFInstance`, `BGPAdvertisement`, the eBPF `vrf_table` entry. `internal/cni/resource.go`'s tracker drops to `vpc, vpcAttachment, vrfCreated, routesCreated` and no longer needs a Kubernetes client.

## hostgw extraction

`ConfigureHostGateway` configures the container's default route — kernel-interface work `galactic-bgp` shouldn't depend on. It now lives in `internal/cni/hostgw`, called directly by both master plugins before handing off down the chain.

## PrevResult vs RawPrevResult

`types.PluginConf.PrevResult` has JSON tag `"-"` and is never populated; only `RawPrevResult map[string]interface{}` is. Pre-existing library quirk, not introduced here — `ops_check.go` already reads `RawPrevResult`, and `inferFromPrevResult` follows the same pattern. Out of scope to fix.

## Verification

- `task lint` ✅ `task build` ✅ `task test:unit` ✅
- `task test:e2e` not run, same caveat as #303/#304.


privateip:
@mattdjenkinson comments address, ready for review

0xmc (dismissed):
lgtm

ecv (dismissed):
Approving. Inferring veth vs tap from sandboxed-interface count rather than `len(Interfaces)` is the right call, and the `cniVersion` constraint being documented rather than buried is appreciated.

Fast-follow: https://github.com/datum-cloud/galactic/issues/331. `galactic-bgp` refuses to run un-chained, but nothing catches the reverse — a conflist missing it gives you a pod with an interface, addresses, and no VPC datapath, all green.

Nit, no issue filed: `cmdDel` runs the full `parseConf`, which can hit `DetectNodeNameFromAPI` and list up to 1000 nodes, purely to log vpc/vpcattachment.

Good to see e2e actually chaining it against real CRDs now.

ecv (approved):

PR #306, opened 2026-08-08
refactor(cni): extract galactic-route as its own CNI chain plugin

privateip:

- #306 refactor(cni): extract galactic-route as its own CNI chain plugin (this PR)

---

## Summary

Fourth branch in the CNI plugin-chain split stack (based on #305). `galactic-route` is now its own chained CNI plugin — invoked between the master plugin and `galactic-bgp` per conflist order — instead of termination-route installation inside `galactic-cni`/`galactic-tap-cni`. Unlike every other binary in the chain, it has zero Kubernetes dependency: no CRD reads or writes, no namespace.

## What moved

`internal/cniroute` mirrors `cniipam`/`cnibgp`'s shape: `RunPlugin()`, `PluginConf{VPC, VPCAttachment, Terminations}`. `parseConf` reuses `config.CNIConfig` for LogFile/LogLevel precedence but never resolves NodeName or Kubeconfig — nothing here talks to the API server.

`cmdAdd` installs each termination as a VRF route via the existing `internal/cni/route` package, deriving the host device name from `(vpc, vpcAttachment)` alone — identical for a veth master's host end and a tap master's tap device, so no interface-kind inference is needed the way `galactic-bgp` needs it. It requires a non-nil `prevResult` and passes it through unchanged.

`cmdDel` is a no-op, matching the pre-split behavior: the old monolithic plugin's own DEL never deleted termination routes either — they're keyed by `(vpc, vpcAttachment)` and may still be in use by another pod/VM, so cleanup stays with `galactic-router`'s GC controller. `cmdCheck` is `checkTerminationRoutes`, moved unchanged. `cmdStatus` is a trivial always-ready success, matching `galactic-ipam`'s own STATUS.

## Rollback scoping

`resourceTracker` covers only the termination routes this plugin's own ADD actually installed (route-delete only) — rollback never touches a route a failed `route.Add` call never reached.

## Master plugin cleanup

`internal/cni`/`internal/cnitap` drop the `Terminations` field from their own `PluginConf` (each had its own copy of a `Termination` type — now lives only in `cniroute`), the `route.Add` loop and `routesCreated` tracker field, and the `checkTerminationRoutes` call from CHECK.

## Verification

- `task lint` ✅ `task build` (all 8 binaries) ✅ `task test:unit` ✅
- `internal/cniroute` at 62.9% coverage — its first-ever coverage, since `internal/cni/route` (the package it wraps) had none before this split either. Backfilling that package's own tests is out of scope here.
- `task test:e2e` not run, same caveat as #303/#304/#305.

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -136,6 +136,7 @@ tasks:
       - go build -ldflags "{{.LDFLAGS}}" -o bin/galactic-tap-cni ./cmd/galactic-tap-cni
       - go build -ldflags "{{.LDFLAGS}}" -o bin/galactic-ipam ./cmd/galactic-ipam
       - go build -ldflags "{{.LDFLAGS}}" -o bin/galactic-bgp ./cmd/galactic-bgp
+      - go build -ldflags "{{.LDFLAGS}}" -o bin/galactic-route ./cmd/galactic-route
       - go build -ldflags "{{.LDFLAGS}}" -o bin/galactic-router ./cmd/galactic-router
       - go build -ldflags "{{.LDFLAGS}}" -o bin/vmtap-cni ./cmd/vmtap-cni
       - GOBIN={{.LOCALBIN}} go install github.com/containernetworking/plugins/plugins/main/host-device@v1.9.1
--- a/cmd/galactic-route/main.go
+++ b/cmd/galactic-route/main.go
@@ -5,7 +5,9 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"strings"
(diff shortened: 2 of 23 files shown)


ecv:
@privateip see matt's comment?

ecv (dismissed):
Approving. Cleanest one in the stack — no Kubernetes, no interface-kind inference, and rollback that only unwinds routes this invocation actually added.

One thing surfaced while reading it, pre-existing rather than yours: https://github.com/datum-cloud/galactic/issues/332. `route.Add` is a bare `netlink.RouteAdd`, so the second pod sharing an attachment hits EEXIST and fails ADD. Same loop existed in the master before, now easier to fix in one place.

Also worth knowing: `terminations` moving onto the `galactic-route` entry means a conflist that leaves them on the master entry installs nothing, silently — same shape as #331.

ecv (dismissed):

ecv (approved):
Re-approving against the current head — the force-push only rebased in two plan docs, no code change since my read.

Findings unchanged: https://github.com/datum-cloud/galactic/issues/332 as a fast-follow, and `terminations` moving onto the `galactic-route` entry means a conflist leaving them on the master installs nothing, silently (#331).

PR #307, opened 2026-08-08
docs+fixes(cni): finish the CNI plugin-chain split — docs, containerlab, dead code

privateip:

- #307 docs+fixes(cni): finish the CNI plugin-chain split - docs, containerlab, dead code (this PR)

---

## Summary

Fifth and final branch in the CNI plugin-chain split stack (based on #306). Docs, containerlab fixtures, and leftover dead code brought up to date with the 5-binary chain steps 0-3 built. No new runtime behavior beyond the fixes below.

`installer.Bootstrap`'s binary-copy list already covers all five chain binaries (added incrementally per step) — nothing left to do there. Confirmed the static host-level conflist only ever carries node-level settings, never per-attachment fields, so its shape needs no change.

## Docs rewritten

- `docs/cni-cmd-sequence.md` — all 3 diagrams rewritten with per-binary lifelines. Also fixes a pre-existing staleness: the diagrams showed a kernel seg6local route install, which the 2026-08-02 eBPF uSID datapath cutover had already replaced with `vrf_table`/`locator_table`/`function_table` map registration.
- `docs/cni/configuration.md` — full rewrite: chain/conflist structure, per-binary field tables, all 7 example configs as multi-plugin conflists. Corrected two inaccuracies: `IPAM.Routes`/`Addresses` are declared but never read by any allocation path; pool IPAM's DEL no longer reads a `BGPAdvertisement` CRD annotation (superseded by on-disk marker files).
- `docs/agents/ARCHITECTURE.md` — repo layout, components, entry points, config, module/package reference, key design decisions, testing, known constraints all updated. Same eBPF-cutover fix as the sequence doc. Added Known Constraints notes for two coverage gaps: e2e doesn't exercise `galactic-route`/`galactic-bgp` (pre-existing, unchanged by the split), and `vmtap-cni` (unrelated binary) has its own doc.

## Real regression found and fixed: containerlab NAD fixtures

All 11 NAD YAML files under `deploy/containerlab/` still used the pre-split single-plugin shape — no `"ipam"` block, no `"plugins"` chain array. Left as-is, every attachment would ADD successfully but silently allocate no IPAM address (top-level fields are unknown JSON to the new `PluginConf`) and never get BGP-advertised (no `galactic-bgp` invocation). Rewrote all 11 as `plugins: [galactic-cni, galactic-bgp]` conflists with addressing moved into an explicit `ipam` block. `deploy/containerlab/docs/tenants.md`'s narrative updated to match.

## e2e test fix

`TestCNITapInterface` invoked `/galactic-cni` directly with `interface_type: tap` and a bogus `srv6_locator` field — both removed in step 0, so it's been exercising nothing since. Fixed to invoke `/galactic-tap-cni` with the explicit-contract `ipam` shape and `CNI_PATH=/` (where the image actually places every binary). Removed the eBPF-control-daemon startup this test no longer needs — eBPF registration is `galactic-bgp`'s job now, which this test doesn't invoke (verifying that end-to-end needs a `BGPRouter` fixture, flagged in Known Constraints, not attempted here — no regression versus the monolithic predecessor's own coverage).

## Dead code removed

- `cmd/galactic-cni/main.go`: deleted the stdin-peek-and-repipe dance and `isTapMode()` — unreachable once `galactic-tap-cni` became its own binary. `cmd/galactic-tap-cni/main.go`'s copy-pasted version simplified too (it already set `CNI_NETNS_OVERRIDE` unconditionally; the buffering around it inspected nothing).
- `internal/config/cni.go`: removed `GALACTIC_CNI_ENABLE_LOCAL_IPAM`/`CNIGetEnableLocalIPAM()` — step 1 introduced the successor but never removed the predecessor; zero production callers remained.
- Doc comments in `internal/plumbing/ebpf/doc.go`, `usidmap/vrf.go`, `internal/plumbing/srv6/usid.go`, `internal/gc/gc.go` still pointing at the deleted `internal/cni/bgp.go` — updated to `internal/cnibgp/bgp.go`.

## Verification

- `task lint` ✅ `task build` (all 8 binaries) ✅ `task test:unit` ✅
- `task test:e2e` not run, same caveat as every prior PR in this stack.

### DIFF
--- a/cmd/galactic-bgp/main.go
+++ b/cmd/galactic-bgp/main.go
@@ -68,11 +68,13 @@ func newRootCommand() *cobra.Command {
 				return nil
 			}
 
-			// Unlike galactic-cni/galactic-tap-cni, this plugin never enters
-			// any network namespace at all (it only makes k8s API calls),
-			// so it needs neither the stdin peek-and-repipe dance nor
-			// CNI_NETNS_OVERRIDE those two use to detect and handle
-			// tap-mode's host-netns invocation.
+			// This plugin never enters a network namespace — it only makes
+			// k8s API calls to publish BGP advertisements. For tap-mode
+			// attachments, CNI_NETNS points at the host netns which equals
+			// this process's ambient netns, so the CNI library's same-netns
+			// rejection check would fire without the override.
+			_ = os.Setenv("CNI_NETNS_OVERRIDE", "true")
+
 			cnibgp.RunPlugin()
 			return nil
 		},
--- a/cmd/galactic-cni/main.go
+++ b/cmd/galactic-cni/main.go
@@ -5,10 +5,8 @@
 package main
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"log"
 	"os"
 	"strings"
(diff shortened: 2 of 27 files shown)


ecv:
hmm red e2e. sec i think i see it

ecv:
eep, sorry for merging on your branch. i think it might help though

ecv (dismissed):
penitence thumb

ecv (commented):
Not approving this one yet — one real break, and this is the PR where it belongs: https://github.com/datum-cloud/galactic/issues/333.

Tap attachments pass the host netns, and `skel` fails ADD for that unless the plugin sets `CNI_NETNS_OVERRIDE` — after the plugin body has already run. `galactic-tap-cni` sets it. `galactic-ipam`, `galactic-bgp` and `galactic-route` don't, so a real tap attach fails with the tap, routes and CRDs already written. `galactic-route` does try, but gates on `interface_type`, deleted back in #303. Three lines each; e2e can't see it because it passes a purpose-made netns rather than the host's.

Everything else here is good — the containerlab NAD rewrite catches exactly the silent-no-IPAM failure #327 is about, and the `isTapMode` and `GALACTIC_CNI_ENABLE_LOCAL_IPAM` removals close out #326.

Happy to approve as soon as the override lands.

ecv:
Tracking what this closes, since it overlaps two things I filed earlier.

#326 — partly. `isTapMode` is gone from `galactic-cni` and the stdin dance from `galactic-tap-cni`. `galactic-route` still has both (that's #333), and `unwrapPathError` is still in `internal/cni`, `internal/cnitap`, `internal/cnibgp` and `internal/cniroute`. Leaving it open for those.

#327 — not closed. The containerlab NADs are fixed, which is the fixture half; the master plugins still accept an old-shape config and attach a pod with no addresses. Worth keeping as the fast-follow.

#325 is untouched here, as expected.

ecv (approved):

ecv (approved):
Approving — #333 is fixed here. All three binaries set `CNI_NETNS_OVERRIDE` unconditionally now, `isTapMode` and the peek-and-repipe are gone from `galactic-route`, and the comments say why rather than restating what.

CI is still running on this head; nothing in the change should trouble it.

I'll close #333 once this merges. #326 and #327 stay open for the parts noted above.

PR #309, opened 2026-08-08
docs: Add implementation plan for VPC ingress EndpointSlice work

### LINKED ISSUE #854: Galactic CNI: publish per-pod SRv6 SID + tenant identifier via EndpointSlice for VPC HTTP ingress backends
## Summary

Per the [HTTP Ingress for VPC Networks design](https://github.com/datum-cloud/enhancements/blob/main/enhancements/networking/http-ingress-for-vpc-networks.md) (PR #851, resolves #853), extend galactic-cni so that every pod attached to a VPC network attachment gets a per-pod `EndpointSlice` carrying its real address plus two annotations — the network attachment's SRv6 Segment Identifier (SID) and its tenant identifier — for the shared, multi-tenant Envoy Gateway ingress path to consume.

**This replaces the previous framing of this issue.** The issue originally described attaching Envoy Gateway pods themselves to a VPC via VPCAttachment. The accepted design does not do that: Envoy Gateway stays on the cluster overlay, and galactic-cni's job is instead to publish routing/tenant metadata for the *VPC-hosted backend pods* that Envoy reaches — not to attach Envoy's own pod to anything.

## Parent Issue

[#796 HTTP Ingress for VPC Networks](https://github.com/datum-cloud/enhancements/issues/796) (via [#853](https://github.com/datum-cloud/enhancements/issues/853)'s accepted design, PR #851)

## Background

Datum's HTTP ingress runs a single shared, multi-tenant Envoy Gateway fleet. Some tenants' `HTTPProxy` backends are pods inside SRv6/EVPN VPCs with tenant-private addressing, where two unrelated tenants can own the identical pod prefix. To let Envoy safely open a connection to the right tenant's pod at that address — without becoming VRF-aware itself and without a dedicated Envoy fleet per tenant — four independent components cooperate:

1. **galactic-cni** (pod's node) — allocates the pod address, computes the network attachment's SID, and publishes both on an `EndpointSlice`
2. **A sidecar** next to Envoy (Envoy's node) — creates the per-tenant Linux VRF device and SRv6 encapsulation route from that EndpointSlice
3. **An Envoy Gateway extension server** (control plane) — patches Envoy's upstream connection config with a socket-bind option naming the tenant's VRF device
4. **An eBPF program** (pod's node) — decapsulates arriving packets and forwards into the right tenant's table

This issue scopes only galactic-cni's piece (#1).

## Scope

galactic-cni's responsibility is the **Setup** and **Teardown** halves of the design, from the pod's-node side:

- **SID computation** — compute one SRv6 Segment Identifier per VPC network attachment (not per pod), using the *existing* SID mechanism already implemented for intra-VPC ingress (`internal/plumbing/srv6/usid.go`, `internal/plumbing/ebpf/uformat/uformat.go`): locator Block, node NodeID, a fixed Function nibble, and the BGPVRFInstance's per-node Argument/VRFID. No change to this mechanism is required — see the "tenant identifier" resolution below. Every pod sharing one network attachment shares the identical SID.
- **EndpointSlice publication** — on pod ADD, publish (or update) a `discovery.k8s.io/v1` `EndpointSlice`, in the pod's own namespace, named after the pod, carrying:
  - the pod's actual allocated address, and
  - the SID and tenant identifier as annotations.

  One `EndpointSlice` per pod; every pod on the same network attachment carries identical SID/tenant-identifier annotation values.
- **Teardown** — on pod DEL, remove or update the pod's `EndpointSlice` **immediately and unconditionally**, same as any other DEL cleanup. Sequencing against the sidecar/extension-server's own unwind (see Open Questions) is explicitly not galactic-cni's concern — the CNI's only obligation is firing an accurate, timely event; any grace period to avoid the teardown race is owned entirely downstream.
- **Tenant identifier** — **resolved, see Open Questions below.** galactic-cni publishes the pod's `(vpc, vpcAttachment)` identifiers — already tracked internally, no new allocation — as the EndpointSlice's tenant-identifier annotation, reusing the same deterministic naming convention `internal/plumbing/intf/intf.go` already uses for kernel interface names.

## Explicitly out of scope (owned by other components in the design)

- Attaching Envoy Gateway's own pod to a VPC network or granting it a VPC-internal IP — **no longer part of the design**; Envoy stays on the cluster overlay.
- Creating/tearing down the per-tenant Linux VRF device or SRv6 encapsulation route on Envoy's node — owned by the sidecar. Per the tenant-identifier resolution below, the sidecar derives its VRF device name from `(vpc, vpcAttachment)` via the same `intf` naming convention, independent of galactic-cni.
- Patching Envoy's upstream connection configuration with the socket-bind option — owned by the Envoy Gateway extension server.
- Decapsulation and tenant-table forwarding lookup on packet arrival — owned by the eBPF program on the pod's node (ownership by a galactic component vs. something else needs confirming).
- **Sequencing/grace-period logic for the teardown race** — owned entirely by the sidecar and extension server (see Open Questions). galactic-cni removes its `EndpointSlice` on DEL the same way it always has; it does not delay, stage, or coordinate that removal on their behalf.
- **Cross-namespace reference resolution between the CNI's `EndpointSlice` and the ingress controller's internal rule representation** — owned by the ingress controller (placement policy and/or an access-grant object are the design doc's two candidate fixes, neither chosen). galactic-cni always publishes the `EndpointSlice` in the pod's own namespace, as it does today for every other pod-scoped resource; it does not change that behavior to accommodate wherever the referencing `HTTPProxy` rule happens to live.

## Acceptance Criteria

- [ ] galactic-cni computes one SRv6 SID per VPC network attachment (shared across all pods on that attachment), not per pod, using the existing SID mechanism unchanged
- [ ] ADD flow publishes a per-pod `EndpointSlice`, in the pod's own namespace, carrying the pod's address, the SID, and a `(vpc, vpcAttachment)`-derived tenant-identifier annotation
- [ ] Multiple pods on the same network attachment produce `EndpointSlice`s with identical SID/tenant-identifier values
- [ ] DEL flow removes/updates the pod's `EndpointSlice` immediately, with no added delay or sequencing logic, so downstream watchers observe pod/attachment removal as soon as it happens
- [ ] CHECK validates the `EndpointSlice`'s address/annotations still match current attachment state
- [ ] GC cleans up stale `EndpointSlice`s left behind by terminated pods
- [ ] Error handling follows CNI spec (DEL idempotent, proper error codes)
- [ ] Unit tests for ADD/DEL/CHECK/GC paths covering SID computation and EndpointSlice publication

## Resolved Questions (inherited from the design doc)

- ~~Where does the "per-tenant identifier" come from — new allocation, or reuse of an existing Galactic/VPC identifier?~~ **Resolved.** No fabric-wide per-tenant integer exists today, and none should be introduced:
  - `BGPVRFInstance.Spec.VRFID` (the SID's 12-bit Argument field, allocated via `allocateArgument()` in `internal/cnibgp/bgp.go`) is documented as "unique per (VPC, PoP)" — scoped per node, not per tenant fabric-wide. The same VPC can get a different Argument value at every PoP it attaches to, so it cannot serve as the single stable value the design implies ties a SID, a VRF device, and a routing table together.
  - The design doc actually conflates two separable concerns. Kernel encapsulation/decap (Setup/Route-lookup steps 1–3) is already fully handled by the existing SID mechanism — no change needed there. VRF device naming on Envoy's node and tenant grouping in the extension server need a stable, collision-free *name*, not a bit-field — which `internal/plumbing/intf/intf.go`'s existing `GenerateInterfaceNameVRF(vpc, vpcAttachment)` (deterministic base62 encoding, already fits the 15-char kernel interface-name limit) already provides.
  - **Decision:** galactic-cni publishes `(vpc, vpcAttachment)` — identifiers it already tracks — as the EndpointSlice's tenant-identifier annotation. The sidecar independently derives its VRF device name from those same two values via the existing `intf` naming convention. No new allocator, no new coordination round-trip.
- ~~Teardown race: no ordering guarantee today between this EndpointSlice update and the sidecar/extension-server unwind — a grace period is discussed but undesigned. Does galactic-cni need to change DEL behavior to narrow this window?~~ **Resolved: not a CNI concern.** galactic-cni's obligation ends at firing an accurate, immediate `EndpointSlice` update on DEL — the same unconditional cleanup behavior it already has for every other DEL-path resource. Any grace period, staged teardown, or sequencing against the sidecar/extension-server's own unwind is entirely their responsibility to design and implement; galactic-cni introduces no new delay or coordination logic to accommodate it. If a future design revision needs the CNI to participate (e.g. a drain signal before removal), that would be a new, separate ask against this issue's scope, not an implicit one.
- ~~Cross-namespace placement of VPC-backed workloads relative to the `HTTPProxy` rule referencing them — unresolved in the design.~~ **Resolved: not a CNI concern.** galactic-cni publishes the `EndpointSlice` in the pod's own namespace, exactly as it already does for every other pod-scoped resource it manages — it does not adapt to, or need visibility into, wherever the ingress controller's internal rule representation lives. Reconciling the two namespaces (via workload placement constraints or a cross-namespace access-grant object — the design doc's two candidate fixes) is entirely the ingress controller's problem to solve, tracked against #796/#851, not this issue.

## Deliverables

- CNI plugin code changes: SID computation + EndpointSlice publish/update/delete on ADD/DEL
- Updated CNI configuration schema documentation reflecting the new EndpointSlice-publishing responsibility
- Unit tests for ADD/DEL/CHECK/GC paths
- Integration test validating EndpointSlice publication and teardown for a VPC-attached pod

## Additional Context

- Full design: [HTTP Ingress for VPC Networks](https://github.com/datum-cloud/enhancements/blob/main/enhancements/networking/http-ingress-for-vpc-networks.md) (PR #851, resolves #853)
- VPC addressing: [Tenant Addressing Plan](architecture/design/network/addressing/tenant.md) — unaffected by this change; SID computation is additive to existing address allocation, not a replacement for it
- CNI spec v1.1.0: Interface plugin with IPAM delegation pattern
- Related: galactic-router distributes EVPN paths via BGP; this issue does not change that path

privateip:
## Summary

datum-cloud/enhancements#854 asks galactic-cni to publish per-pod routing metadata so the shared Envoy Gateway ingress fleet can reach VPC-hosted backends, per the design accepted in datum-cloud/enhancements#851. This adds a written implementation plan mapping that work onto the CNI's current plugin-chain architecture before any code is written. It also resolves four engineering questions the design left open: address-family scope, a new discovery label for a not-yet-built downstream component, service-account ownership, and rollback behavior on partial failure.

## Test plan

- [ ] Plan doc renders cleanly on GitHub (headings, tables, links)

Related to datum-cloud/enhancements#854

### DIFF
--- a/docs/plans/854-vpc-http-ingress-endpointslice.md
+++ b/docs/plans/854-vpc-http-ingress-endpointslice.md
@@ -1,0 +1,98 @@
+# Implementation Plan — #854: EndpointSlice publication for VPC HTTP ingress
+
+- **Issue:** [datum-cloud/enhancements#854](https://github.com/datum-cloud/enhancements/issues/854)
+- **Parent:** [datum-cloud/enhancements#796](https://github.com/datum-cloud/enhancements/issues/796) — HTTP Ingress for VPC Networks
+- **Design doc:** [HTTP Ingress for VPC Networks](https://github.com/datum-cloud/enhancements/blob/main/enhancements/networking/http-ingress-for-vpc-networks.md) (PR [#851](https://github.com/datum-cloud/enhancements/pull/851), resolves [#853](https://github.com/datum-cloud/enhancements/issues/853))
+- **Status:** planning only — no implementation started
+
+## Correction to #854's framing
+
+Galactic-cni has already been split into a chained-plugin architecture (commits `c1704f7`/`e33092c`/`725ef87`). `internal/cni` (the master plugin) no longer talks to Kubernetes for BGP state at all — that moved to `internal/cnibgp`, the **last plugin in the chain** (`galactic-cni` → `galactic-ipam` → `galactic-route` → `galactic-bgp`). #854 was written before/around that split and refers to "galactic-cni" generically. All of this work belongs in **`internal/cnibgp`**, because it's the only plugin in the chain with all four required inputs simultaneously in scope: the pod's real allocated address (from `prevResult`), `vpc`/`vpcAttachment`, the SRv6 locator/nodeID (from `BGPRouter`), and the allocated VRFID/Argument.
+
+## 1. Scope recap (resolved state of #854)
+
+- Compute the SID using the **existing, unmodified** mechanism (`srv6.ComputeSID`) — no new SID logic.
+- Publish one `discoveryv1.EndpointSlice` per pod, named after the pod, in the pod's own namespace, carrying the pod's address + SID annotation + `(vpc, vpcAttachment)`-derived tenant annotation.
+- Remove it immediately and unconditionally on DEL — no grace period, no sequencing with sidecar/extension-server (not a CNI concern).
+- CHECK validates it; GC cleans up stragglers.
+- Not in scope: Envoy VPC attachment, sidecar VRF/route, extension-server config patching, eBPF decap, cross-namespace `HTTPProxy` reference resolution.
+
+## 2. Where each piece lands
+
+| Concern | Binary | Package | Existing anchor |
+|---|---|---|---|
+| SID computation | `galactic-bgp` | `internal/cnibgp` | `publishBGPState`, `bgp.go:318-408` |
+| EndpointSlice publish (ADD) | `galactic-bgp` | `internal/cnibgp` (new file) | after `bgp.go:373-398` (BGPAdvertisement CreateOrUpdate) |
+| EndpointSlice delete (DEL) | `galactic-bgp` | `internal/cnibgp/ops_del.go` | currently a 6-line no-op (`ops_del.go:22-28`) |
+| EndpointSlice validate (CHECK) | `galactic-bgp` | `internal/cnibgp/ops_check.go` | after BGPAdvertisement Get (`ops_check.go:64-68`) |
+| Naming/annotation vocabulary | both | `internal/cni/crdnames` | already the shared cross-plugin vocab package |
+| Pod-name parsing | `galactic-bgp` (new dependency) | `internal/cni/nadpatch` | sibling to existing `ParsePodNamespace` (`nadpatch.go:40-48`) |
+| RBAC (publish/delete) | `galactic-bgp`'s SA | `config/cni/rbac.yaml` | shares the `galactic-cni` SA (`daemonset.yaml:17`) — confirmed |
+| GC backstop | `galactic-router` | `internal/gc/gc.go`, `internal/controller/gc_controller.go` | extend `RemoveOrphanedCRDs` (`gc.go:189-231`), not the CNI's eBPF sweep |
+| GC RBAC | `galactic-router`'s SA | `config/router/rbac.yaml` | mirror existing `bgpadvertisements` delete grant (`rbac.yaml:15-19`) |
+
+## 3. Work items, in dependency order
+
+### Phase 1 — Shared vocabulary (`internal/cni/crdnames`)
+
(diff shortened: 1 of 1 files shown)


privateip:
The failing CI check is due to #302 not being merged yet.  Once it is merged, I will rebase this PR to clean up the CI failures.

ecv (approved):

PR #311, opened 2026-08-08
fix(vrf): key kernel VRF by VPC, not (VPC, attachment)

privateip:
## Summary

Two pods attaching to the same VPC on the same node collided: the kernel VRF interface was keyed per attachment, so a second attachment's setup saw the first's still-live VRF, mistook it for leftover state from a failed attempt, and deleted it, silently breaking the first pod's networking. The VRF is now shared per VPC per node, matching how a VPC is a per-node routing domain, not a private resource of one attachment.

## Test plan

- [ ] `task ci` passes (lint, build, unit tests)
- [ ] Two attachments sharing a VPC on one node no longer collide on VRF teardown

---

- #311 fix(vrf): key kernel VRF by VPC, not (VPC, attachment) (this PR)
- #312 fix(cnibgp): share BGPVRFInstance and Argument per (VPC, node)
- #313 fix(containerlab): give ns30/ns40 two attachments, not one shared

### DIFF
--- a/internal/cni/cni_test.go
+++ b/internal/cni/cni_test.go
@@ -822,9 +822,6 @@ func TestResourceTrackerFieldsSet(t *testing.T) {
 	if tracker.vpcAttachment != testAttachment {
 		t.Errorf("vpcAttachment = %q, want %q", tracker.vpcAttachment, testAttachment)
 	}
-	if tracker.vrfCreated {
-		t.Error("vrfCreated should be false by default")
-	}
 }
 
 // ---- cmdStatus ---------------------------------------------------------
--- a/internal/cni/hostgw/hostgw.go
+++ b/internal/cni/hostgw/hostgw.go
@@ -33,12 +33,28 @@ import (
 )
 
 // ConfigureHostGateway assigns each configured family's gateway address as a
-// host address (/128 for IPv6, /32 for IPv4 on veth) on the host-side
-// interface (veth or tap) and installs an explicit pod-subnet route for that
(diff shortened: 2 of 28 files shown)


ecv (approved):
Approving. Dropping the VRF from both masters' rollback is the right call — `vrf.Add` being idempotent means a `vrfCreated` flag could never tell "I made it" from "a sibling already had" — and the GC rewrite handles the shared case properly, including treating an advertisement it can't judge as surviving rather than guessing.

The NDP note in `hostgw` is worth the space. Gateway on the VRF device looks obviously right until you find out proxy_ndp doesn't behave like proxy_arp.

Fast-follow: https://github.com/datum-cloud/galactic/issues/334. VRFs created before the rename don't match the new pattern, so GC skips them entirely and they strand on any in-place upgrade.

Worth stating for whoever merges: don't ship this without #312. The kernel table is shared per VPC here, but `BGPVRFInstance` is still `vpc-vpcAttachment`, so two attachments on one VPC each get their own Argument and both write the same shared `vrf_table` slot — last writer wins.

Also widened #332: termination prefixes from different attachments now land in one table, so overlapping ones collide across attachments too.

ecv:
Merge-order note: #311 and #312 need to land together.

#311 shares the kernel VRF per VPC, but `BGPVRFInstance` stays `vpc-vpcAttachment` until #312 — so two attachments on one VPC get separate Arguments and both write the same shared `vrf_table` slot. Last writer wins.

PR #312, opened 2026-08-08
fix(cnibgp): share BGPVRFInstance and Argument per (VPC, node)

privateip:
## Summary

The BGPVRFInstance CRD and its SRv6 Argument were still allocated per attachment instead of per VPC per node, wasting an Argument slot for every extra attachment sharing a VPC. This makes every attachment on the same VPC/node converge on one CRD and Argument, and fixes rollback so a failed attachment's cleanup can no longer delete a live sibling's shared state.

## Test plan

- [ ] `task ci` passes (lint, build, unit tests)
- [ ] A failed ADD's rollback never deletes a still-live sibling attachment's BGPVRFInstance
- [ ] Two different VPCs' first attachment landing on the same node at the same time no longer wedges permanently on an argument collision

---

- #311 fix(vrf): key kernel VRF by VPC, not (VPC, attachment)
- #312 fix(cnibgp): share BGPVRFInstance and Argument per (VPC, node) (this PR)
- #313 fix(containerlab): give ns30/ns40 two attachments, not one shared

### DIFF
--- a/internal/cni/crdnames/crdnames.go
+++ b/internal/cni/crdnames/crdnames.go
@@ -69,10 +69,16 @@ func NetNSKey(containerID string) string {
 }
 
 // BGPVRFInstanceName returns the deterministic name for a BGPVRFInstance.
-// Each VPCAttachment is unique per interface across the cluster, so the
-// (vpc, vpcAttachment) pair is a reliable 1:1 key.
-func BGPVRFInstanceName(vpc, vpcAttachment string) string {
-	return fmt.Sprintf("%s-%s", vpc, vpcAttachment)
+// Unlike BGPAdvertisementName, this is keyed by (vpc, node) rather than
+// (vpc, vpcAttachment): the underlying kernel VRF is shared by every
+// attachment (pod or VM) on this VPC on this node (see internal/plumbing/vrf
+// and internal/plumbing/intf's GenerateInterfaceNameVRF), so every attachment
+// on the same VPC/node must resolve to the same BGPVRFInstance rather than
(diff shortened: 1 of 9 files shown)


ecv:
Merge-order note: #311 and #312 need to land together.

#311 shares the kernel VRF per VPC, but `BGPVRFInstance` stays `vpc-vpcAttachment` until #312 — so two attachments on one VPC get separate Arguments and both write the same shared `vrf_table` slot. Last writer wins.

ecv (approved):
Approving — this is careful work, and it reads that way.

Two decisions I want to call out as right, since both look like omissions until you follow the reasoning. Gating `vrfInstanceCreated` on `OperationResultCreated` rather than "CreateOrUpdate succeeded" is what makes the argument-collision retry heal instead of wedge. And dropping the eBPF rollback entirely is the correct read of that state's lifetime — `SweepEBPFVRFTable` already reclaims it, with a generation cutoff and a no-BGPRouter guard that stops a bad tick from wiping the node. Net 190 lines lighter for a behavioural fix is a good trade.

One fast-follow, no need to hold the stack: https://github.com/datum-cloud/galactic/issues/335. "I created it" and "nobody else has adopted it yet" come apart when two attachments on one VPC race on a node — a sibling can pick up the freshly created BGPVRFInstance before this ADD fails, and rollback then removes it out from under a live attachment. Narrow window, and it needs exactly the concurrency this PR makes ordinary. A guard on any other BGPAdvertisement for the VPC and node would keep the collision self-heal intact.

CI is green here, E2E included.

PR #313, opened 2026-08-08
fix(containerlab): give ns30/ns40 two attachments, not one shared

privateip:
## Summary

ns30/ns40 modeled "two pods, one VPC, one node" as one NAD scaled to two replicas, sharing a single attachment identifier — the setup that triggered the VRF collision fixed earlier in this stack. Each now gets its own NAD and Deployment with a distinct attachment under the same VPC, matching how every other tenant fixture models one attachment as one pod.

## Test plan

- [ ] `task verify` passes end-to-end on a live containerlab lab, including ns30/ns40's two-attachment same-node pings
- [ ] ns30's two pods are confirmed sharing one kernel VRF across distinct attachments

---

- #311 fix(vrf): key kernel VRF by VPC, not (VPC, attachment)
- #312 fix(cnibgp): share BGPVRFInstance and Argument per (VPC, node)
- #313 fix(containerlab): give ns30/ns40 two attachments, not one shared (this PR)

### DIFF
--- a/deploy/containerlab/README.md
+++ b/deploy/containerlab/README.md
@@ -138,12 +138,15 @@ deploy/containerlab/
 │   ├── galactic-router/         # galactic-router DaemonSet + BGP CRs (dfw, iad, sjc)
 │   ├── galactic-control/iad/    # galactic-router RR + BGP CRs (iad-control)
 │   └── tenants/                 # test VPCs — one shared base/ (Namespace + netshoot
-│       ├── base/                 # Deployment), each tenant patching its namespace,
-│       ├── ns50/                 # default-network annotation, and (ns30/ns40) replica
-│       ├── ns10/                 # count; per-site dirs hold each site's NAD(s):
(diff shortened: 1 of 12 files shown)


ecv (approved):
Approving. Two distinct attachments under one VPC is the right shape for this fixture — the old "one NAD, replicas: 2" version was reproducing the veth-name collision rather than the case the stack set out to prove, and the doc comments explaining why are the kind that stop someone reverting it in six months.

Nice catch on `iad-worker-control` in the troubleshooting section: a tainted worker that the control-plane affinity does not exclude is exactly the thing someone would debug twice.

Fast-follow: https://github.com/datum-cloud/galactic/issues/336. Same-node placement is what these two fixtures test, and it currently holds only because dfw has one worker and iad's second is tainted — nothing in the fixtures or the verify scripts checks it. If that shape ever changes the pods spread, the pings still pass cross-node, and the test quietly stops testing its own subject. A `.spec.nodeName` comparison in each script covers it.

Build, lint and both unit jobs are green; E2E still running.

PR #316, opened 2026-08-08
refactor(cni): dedupe internal/cni vs internal/cnitap into internal/cnimaster

privateip:
Follow-up to mattdjenkinson's [review on #303](https://github.com/datum-cloud/galactic/pull/303#issuecomment-5226782951) — the last of the review's 5 points (the other 4: 2 already resolved upstream in the stack, 1 now moot, 2 landed in #(fix/cni-review-followups-doc-placement), which this branch stacks on).

## Problem

`internal/cnitap` reimplemented `internal/cni`'s `config.go` (`parseConf`'s validation, log setup, status parsing, prevResult validation, base62/error sanitizing), `resource.go` (`newK8sClient`, the tracker/cleanup rollback sequence), and `ops_check.go` (`checkNodeLevelState`, `cmdStatus`, `probeAPIServer`, `validateHostInterface`) almost verbatim — none of it is interface-specific. A fix to one wouldn't have propagated to the other. `internal/cnitap`'s own tests were implicitly relying on `internal/cni`'s tests to cover the logic it silently duplicated, which is exactly the risk the review comment flagged.

## Fix

Extracted the interface-agnostic pieces into a new `internal/cnimaster` package — the same kind of shared-library extraction this stack already did for hostconf/crdnames/nadpatch/cniipam/cnibgp:

- `PluginConf` moves to `cnimaster`; `internal/cni` and `internal/cnitap` each keep `type PluginConf = cnimaster.PluginConf` (mirroring the existing `HostConf = hostconf.HostConf` alias pattern).
- `ParseConf`, `LoadHostConf`, `ParseStatusConf`, `ValidatePrevResult(Add)`, `IsValidBase62`, `SanitizeForError`, `UnwrapPathError`, `ParseLogLevel`, `SetupLogging` move to `cnimaster/config.go`. Each package's own `parseConf` becomes a two-line wrapper binding `ParseConf` to its own `cniConfig`/`ConfFile` (these stay per-binary — each resolves its own `GALACTIC_CNI_*` environment independently).
- `NewK8sClient` (+ the backing scheme) and a new `CleanupAttachment` (parameterized by interface kind and a `Delete` func, replacing the copy-pasted rollback sequence) move to `cnimaster/resource.go`.
- `CheckNodeLevelState`, `ValidateHostInterface`, `ProbeAPIServer(Fn)`, and a new `RunStatus` (the full `cmdStatus` body, since it wasn't interface-specific either) move to `cnimaster/check.go`. Each package's own `cmdStatus` becomes a one-line wrapper.
- `cmdCheck`'s guest-side validation (`checkGuestInterface`, `validateGuestInterface`, `validateIPOnInterface`) stays in `internal/cni` only — tap has no guest netns to check, so this piece is genuinely interface-specific and wasn't touched.

Test coverage moved with the code: the generic-logic unit tests (`TestParseConf`, `TestIsValidBase62`, `TestSanitizeForError`, `TestValidatePrevResult(Add)`, `TestProbeAPIServer*`, `TestLoadHostConf`, `TestParseLogLevel`, `TestLoggingSetup*`, `TestIPAMBlockPresenceIsTheOnlyTrigger`) moved from `internal/cni/cni_test.go` into `internal/cnimaster/cnimaster_test.go`. `internal/cnitap`'s own `TestLoadHostConfMissingFile` (testing a function that no longer exists in that package) is removed as now fully redundant with `cnimaster`'s own coverage. Remaining tests in both `cni_test.go`/`cnitap_test.go` are either interface-specific (`cmdAdd`/`cmdDel`/`cmdCheck`/`buildResult`) or thin wiring smoke tests over the shared functions.

Net effect: `internal/cni` and `internal/cnitap` together lose ~750 lines of near-duplicate implementation+test code; `internal/cnimaster` adds ~460 (including its own full test suite) — about 250 fewer lines overall, in one place instead of two.

## Verification

- `task lint` ✅
- `task build` ✅ (all 8 binaries)
- `go vet ./...` ✅
- `task test:unit` ✅ (33/33 packages passing, 0 failures)
- `task test:e2e` not run (same caveat as every step in this stack — needs a Kind cluster bring-up not attempted here)


ecv (approved):
Approving — this is the fix for #325, and the split lands where I'd have put it. `CleanupAttachment` taking the delete func keeps the shared-VRF decision from #311 intact (nothing in either master deletes the VRF any more), IPAM release still runs before interface teardown in both, and `RejectMovedIPAMKeys` survives in the shared `ParseConf`. Guest-netns validation staying in `internal/cni` is the right line to draw.

One thing to squash in before merge, three lines: `cnimaster.LoadHostConf` uses `os.IsNotExist(UnwrapPathError(err))`, which re-introduces the hand-rolled unwrap #339 deleted four hours ago. `hostconf.Load` wraps with `%w`, so `errors.Is(err, fs.ErrNotExist)` works and `UnwrapPathError` can go. As it stands `galactic-bgp` and `galactic-route` use the standard check while both masters go back to the helper — the shared package ends up carrying the older shape. Easy to miss in a rebase this size; happy for it to be a follow-up if you'd rather not touch the branch.

Verified against main rather than taking the description's word: node-name detection, namespace defaulting, base62 validation, logging setup and the KUBECONFIG side effect all appear exactly once in the shared `ParseConf`, same as the original.

CI green including E2E. I'll close #325 once this merges.

Issue #336, opened 2026-08-10
ns30/ns40 rely on cluster shape to stay same-node, and never check


ecv:
ns30 and ns40 exist to prove two attachments on one VPC share a VRF on a single node. Nothing in either fixture or its verify script says the two pods must land on the same node.

Today they do, for reasons outside the fixtures: dfw has exactly one worker, and iad's second worker carries the route-reflector taint. Remove that taint, add a worker, or reuse the pattern in a cluster shaped differently, and the pods spread — the pings still pass over the cross-node path, and the case the fixture is named for stops being exercised with no signal that it stopped.

## Outcome

The verify scripts assert both pods are on one node, so the test fails loudly when the property it depends on goes away.

Related to #313

ecv:
What holds it together today, on #313's head:

- dfw (`node_files/dfw/config.yaml`) has one worker, so ns30 has nowhere else to go.
- iad has two, and `node_files/iad/config.yaml` taints `iad-worker-control` `NoSchedule` for the route-reflector role — the only thing keeping ns40's pair together, since `pod.yaml`'s affinity excludes Kubernetes control-plane nodes and that one is a worker. `docs/tenants.md` already flags this.

`verify-ns30.sh` and `verify-ns40.sh` resolve each pod by label through `pod_name` in `scripts/lib.sh`; the `NODE` argument there is the docker container kubectl runs in, not a scheduling constraint. Neither script reads `.spec.nodeName`.

A `jsonpath='{.spec.nodeName}'` comparison between the two pods, failing when they differ, is enough.

ecv:
Closed by #337 (merged `332414d`): `verify-ns30.sh` and `verify-ns40.sh` compare both pods' `.spec.nodeName` and fail with the node each landed on when they split.

PR #337, opened 2026-08-10
test(containerlab): assert same-node VPC fixtures actually stay same-node

ecv:
## Summary

Two of the lab's tenant VPCs exist to prove one specific thing: when a single VPC has two attachments on the same worker, both share one kernel VRF and talk to each other locally, with no cross-site hop involved.

Nothing in either fixture actually held the two workloads on one worker — that was a side effect of how the lab clusters happen to be shaped today, and reshaping a cluster would quietly scatter them.

Scattered workloads still ping each other successfully over the cross-site path, so the checks would keep passing while no longer testing the thing they are named for.

Both checks now compare where the two workloads were placed and fail with a clear message naming each one's location when they end up apart, so the day that property goes away the lab says so instead of going green.

## Test plan

- Shell syntax check on every touched script passed.
- Static shell analysis on the same scripts is clean.
- The real check is a live lab run of both scenario verifications, once against the current lab and once with the placement property deliberately broken to confirm the new failure fires. That has not been done — a lab was not available in this environment.

Related to #336

### DIFF
--- a/deploy/containerlab/docs/tenants.md
+++ b/deploy/containerlab/docs/tenants.md
@@ -24,8 +24,8 @@ They differ only in scope and addressing:
 | `ns50` | dfw, sjc, iad (3-site)       | IPv4 + public IPv6 ptp | `G000000050V` | Also defines a `public` NAD with an IPv6 IPAM pool for external-connectivity testing. |
 | `ns10` | dfw, sjc, iad (3-site)       | IPv6-only (fd20 ULA) | `G000000010V` | No `ipv4_subnet` at all. |
 | `ns20` | dfw, sjc, iad (3-site)       | Dual-stack (fd20 ULA + IPv4) | `G000000020V` | Both families active; exercises the dual-stack IPAM path. |
-| `ns30` | dfw only, 2 attachments      | IPv6-only (fd20 ULA) | `G000000030V` | Two distinct attachments (`private`/`private-b`, distinct `vpcattachment` values, same `vpc`), each its own single-replica Deployment, both land on `dfw-worker` and share one VRF — same-node connectivity, no cross-site hop. |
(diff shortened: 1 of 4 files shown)


ecv:
File-level detail.

`scripts/lib.sh` gains `pod_scheduling_node NODE NAMESPACE POD`, which reads `.spec.nodeName` via jsonpath. It sits with the other `pod_*` helpers. Its `NODE` argument means what it means everywhere else in that file — the docker container `kubectl` runs in — so the name carries the distinction from the value it returns.

`verify-ns30.sh` and `verify-ns40.sh` resolve both pods' scheduling nodes after the existing 2-pod count check and before address resolution, and exit 1 to stderr naming each pod and its node when the two differ. On success they print the shared node. The comment above each block records why the pods sit together today, which differs per site: dfw has one schedulable worker, iad relies on the route-reflector taint.

`docs/tenants.md` picks up the new assertion in the VPC table, in the connectivity-check section alongside the helper list, and in the `iad-worker-control` troubleshooting entry, which now names the failure text you would see if that taint disappeared.

Verified: `bash -n` on all three scripts, and `shellcheck -x` clean across them. No live lab run, as noted in the test plan.

ecv:
Merge-order note: this touches the same lines of `verify-ns30.sh`/`verify-ns40.sh` and the same `tenants.md` rows as #313, which is approved and not yet merged.

#313 replaces the two-pod `pod_names` lookup with per-label lookups, and this inserts the node comparison right below that block — so whichever lands second needs a rebase. Suggest merging #313 first; I'll rebase this on top and re-request review.

ecv:
Correction to the note above: #313 had already merged. Rebased onto current main — the ns30/ns40 label-based pod lookup and the updated VRF names in the docs table are both preserved, and the node assertion now sits after the new lookup. No conflict.

privateip (approved):

Issue #334, opened 2026-08-10
Pre-rename VRFs are invisible to the garbage collector


ecv:
Kernel VRF interfaces are now named after the VPC alone. Every VRF created before that change carries the older name with an attachment segment, and the garbage collector's pattern no longer matches it.

Anything that does not match is treated as "not ours" and skipped, so on a node upgraded in place the old VRFs are never collected. They keep their routing table IDs and whatever routes were in them, indefinitely, with nothing left that refers to them.

Pods already running on those VRFs keep working until they restart, at which point they land on a new VRF and the old one is stranded.

## Outcome

Either the collector recognises and reclaims the old shape, or draining before upgrade becomes a documented step.

Related to #311

ecv:
`internal/gc/gc.go`'s `vrfNameRegex` is now `^G([A-Za-z0-9]{9})V$` (11 chars). Anything created before #311 matches the old `^G([A-Za-z0-9]{9})([A-Za-z0-9]{3})V$` (14 chars).

`CollectOrphanedVRFs` skips non-matching names outright ("Not a Galactic VRF"), so a legacy VRF never reaches `RemoveOrphanedVRFs` — whose by-name fallback would have handled it fine had it been collected.

Cheapest fix: keep the old pattern as a second alternative in collection only, matching it to the same VPC, so pre-rename VRFs get reclaimed once nothing references their VPC.

ecv:
Closed by #338 (merged `eb59f20`): `CollectOrphanedVRFs` resolves both name shapes via `vpcFromVRFName`, and a legacy interface is deleted by the name actually observed rather than one rebuilt from the VPC.

PR #338, opened 2026-08-10
fix(gc): reclaim VRFs created before the per-VPC rename

ecv:
Galactic's per-VPC VRFs were renamed a while back, and garbage collection only ever learned the new name. Any VRF a node created before that rename is invisible to GC — it is skipped as "not ours" and survives forever, quietly holding a routing table ID and its routes long after every workload that needed it is gone. This only bites nodes upgraded in place rather than rebuilt, which is exactly where it is hardest to notice.

Garbage collection now recognises the older name too, resolves it to the same VPC, and reclaims it once nothing on the node still references that VPC.

## Test plan

- New unit tests cover the older name resolving to the right VPC, the current name continuing to resolve, and non-Galactic interface names continuing to be ignored.
- Existing garbage collection unit tests unchanged and still passing.
- No change to how names are generated — only to what cleanup is willing to recognise.

Related to #334

### DIFF
--- a/internal/gc/gc.go
+++ b/internal/gc/gc.go
@@ -56,6 +56,19 @@ type CleanupResult struct {
 // Base62 includes digits and letters.
 var vrfNameRegex = regexp.MustCompile(`^G([A-Za-z0-9]{9})V$`)
 
+// legacyVRFNameRegex matches the VRF interface name Galactic generated before
+// the VRF became per-VPC: the template was "G%09s%03sV", carrying the same
+// VPCAttachment segment the host/guest veth names still do (14 chars). A node
+// upgraded in place keeps whatever VRFs it created under the old template, and
+// those interfaces hold a routing table ID and its routes for as long as they
+// survive — so collection has to recognise them, map them to the same VPC the
+// current name would yield, and reclaim them once nothing references that VPC.
+//
+// This is deliberately a collection-only concern: nothing creates these names
+// any more, and the removal path must NOT resolve one back through
+// intf.GenerateInterfaceNameVRF (see vpcFromVRFName and RemoveOrphanedVRFs).
+var legacyVRFNameRegex = regexp.MustCompile(`^G([A-Za-z0-9]{9})[A-Za-z0-9]{3}V$`)
+
 // routerNamesForNode returns the names of every BGPRouter in the namespace
 // whose TargetRef points at nodeName. BGPAdvertisement/BGPVRFInstance CRDs
 // are namespace-scoped, not node-scoped — a namespace can hold CRDs created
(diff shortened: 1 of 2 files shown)


ecv:
## Implementation detail

`internal/gc/gc.go` gains `legacyVRFNameRegex` (`^G([A-Za-z0-9]{9})[A-Za-z0-9]{3}V$`) alongside the existing `vrfNameRegex`, and a `vpcFromVRFName` helper that tries the current shape first and falls back to the legacy one, stripping the `%09s` padding either way so the VPC matches CRD naming. `CollectOrphanedVRFs` switches to that helper.

## Why the removal path was left on `parseVRFName`

`RemoveOrphanedVRFs` has two paths. The primary one resolves the name to a VPC and calls `vrf.Delete(vpc)`, which rebuilds the *current* `G%09sV` name via `intf.GenerateInterfaceNameVRF`. Handing it a legacy-derived VPC would look up a name that does not exist on the box; `vrf.Delete` is idempotent, so it returns nil and the pass reports a removal that never happened.

So `parseVRFName` deliberately keeps rejecting the legacy shape. That drops a legacy name into the second path — `netlink.LinkByName` + `LinkDel` on the name actually observed — which deletes the right interface. There is a test case asserting that rejection, with a comment explaining it is load-bearing rather than an oversight.

That fallback also never incremented `OrphanedVRFsRemoved` or logged on success, so it is now the only path a legacy VRF takes and would have under-reported every one of them. Fixed in the same commit.

Route flushing: `vrf.Delete` flushes the table before dropping the link, the fallback does not. Not a leak — `vrf.Add` flushes any table ID it allocates before creating a VRF on it, so a reused ID starts clean either way. Reaching into `internal/plumbing/vrf` for an exported flush would have taken this outside `internal/gc`.

## Verification

- `gofmt -l` on both changed files: clean.
- The package does not build on this macOS host: `internal/plumbing/ebpf/prog`'s bpf2go artifacts are not committed and `llvm-strip` is unavailable locally, so `go test ./internal/gc/...` fails with `undefined: prog.Usid*`. Pre-existing and unrelated to this change — `GOOS=linux GOARCH=amd64 go vet ./internal/gc/` reports errors only in the eBPF packages, none in `internal/gc` itself. Relying on CI for the real run.
- To get actual signal in the meantime, the parsing functions and their tests were extracted verbatim into a throwaway package with no netlink dependency and run there: `TestParseVRFName`, `TestVPCFromVRFName`, and `TestVRFNameRegex` all pass, 24 subtests.

ecv:
One boundary worth recording, since it is safe today for a reason that is not visible in the code.

The legacy pattern matches 14-character names: `G` + 9 + 3 + `V`. A current-shape name for a VPC whose base62 form is 12 characters is also 14 characters, and would be read as a legacy name with a truncated VPC — collected against the wrong VPC, then deleted by name.

That needs a VPC identifier above 64 bits: base62 of a uint64 tops out at 11 characters, which yields a 13-character name and stays unambiguous. It also has to fit the kernel's 15-character interface-name limit, which is why these identifiers are base62 in the first place.

So this is correct as written. If VPC identifiers ever widen past 64 bits, this is the second thing that breaks after the name length itself.

privateip (approved):

Issue #326, opened 2026-08-10
Three leftovers from the CNI chain split


ecv:
Three small leftovers from the CNI chain split. None change behavior today; each is a trap for the next person.

The interface type crosses a package boundary as a bare string, while the package receiving it keeps its own unexported constants for the same two values. A typo picks the wrong eBPF egress kind and nothing complains.

Both master plugins carry a helper that walks an error chain by hand to spot a missing file. The library they call already wraps it so the standard check works directly.

The tap plugin still buffers all of stdin through a pipe before running. That only existed so the old binary could peek at the config to decide which mode it was in. It has no mode to decide anymore.

Related to #303

ecv:
Specifics as of #303's head.

**Interface type as a bare string.** `internal/cni/ops_add.go` and `internal/cnitap/ops_add.go` both build `cnibgp.PublishConfig{... InterfaceType: "veth"}` / `"tap"`. `internal/cnibgp/bgp.go` keeps `ifaceTypeVeth`/`ifaceTypeTap` unexported and feeds the value to `EgressKindForInterfaceType`. Export them, or make it a typed value.

**Hand-rolled error unwrap.** `unwrapPathError` in `internal/cni/config.go` and `internal/cnitap/config.go` walks to the innermost error so `os.IsNotExist` can see it. `hostconf.Load` wraps with `%w`, so `errors.Is(err, fs.ErrNotExist)` works as-is — its own doc comment says so.

**Stdin re-pipe.** `cmd/galactic-tap-cni/main.go` reads all of stdin, writes it back through an `os.Pipe`, and swaps `os.Stdin`. That existed in `cmd/galactic-cni/main.go` to peek `interface_type` before running. The tap binary sets `CNI_NETNS_OVERRIDE` unconditionally and can call `RunPlugin()` on the real stdin.

ecv:
#339 closes the second item — the hand-rolled error-chain walk. It went to four plugins in the end, not two: `internal/cni`, `internal/cnitap`, `internal/cnibgp`, and `internal/cniroute` each carried the same copy. All four now call `errors.Is(err, fs.ErrNotExist)` directly, which is what `hostconf.Load`'s doc comment already promised would work.

The other two items are already handled elsewhere.

The bare interface-type string was resolved when interface kind moved to prevResult inference — the receiving package no longer takes the value as a string across the boundary.

[...]

PR #339, opened 2026-08-10
chore(cni): drop the duplicated missing-conflist unwrap helper

ecv:
## Summary

Four of the CNI plugins each carried a private copy of the same twelve-line helper, written around a limitation the standard library's missing-file check no longer has.

The loader they call already wraps its errors the modern way, so the one-line check recognizes a missing file on its own.

This deletes all four copies and uses that one-liner.

Runtime behavior holds: a node without the shared config file still falls back to defaults, and every other read failure still surfaces.

## Test plan

Formatting and static analysis pass clean over the affected packages, cross-compiled for Linux since they are Linux-only. Existing unit tests already cover the missing-file fallback and run in CI on this branch.

Related to #326

### DIFF
--- a/internal/cni/config.go
+++ b/internal/cni/config.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/fs"
 	"log/slog"
 	"os"
 	"path/filepath"
@@ -71,7 +72,7 @@ func loadHostConf(filePath string) (*HostConf, error) {
 	}
 	conf, err := hostconf.Load(filePath, hostconf.PluginType)
 	if err != nil {
-		if os.IsNotExist(unwrapPathError(err)) {
+		if errors.Is(err, fs.ErrNotExist) {
 			return &HostConf{Namespace: config.DefaultNamespace}, nil
 		}
 		return nil, err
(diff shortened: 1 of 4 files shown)


ecv:
File-level detail for reviewers.

The helper was `unwrapPathError`, duplicated verbatim in `internal/cni/config.go`, `internal/cnitap/config.go`, `internal/cnibgp/config.go`, and `internal/cniroute/config.go`. Each one looped `errors.Unwrap` to the innermost error so `os.IsNotExist` — which does not traverse `%w` — could see the `os.ReadFile` failure underneath.

`hostconf.Load` wraps that failure as `fmt.Errorf("read conflist file %q: %w", filePath, err)`, and its own doc comment already promises the result is checkable with `errors.Is`. So each call site becomes `errors.Is(err, fs.ErrNotExist)` and each package picks up `io/fs`.

`os` and `errors` both stay imported in all four files — `errors.Is` keeps `errors` live, and `os` is still used for `MkdirAll`/`OpenFile` in the log setup.

The three fallback returns differ and are preserved as-is: `cni`, `cnitap`, and `cnibgp` return `&HostConf{Namespace: config.DefaultNamespace}`; `cniroute` returns a bare `&HostConf{}` because it has no namespace default.

No test referenced the helper.

### What I ran

- `gofmt -l` on all four changed files: empty.
- `GOOS=linux GOARCH=amd64 go vet` clean on `./internal/cni/...`, `./internal/cnitap/...`, `./internal/cniroute/...`.
- `internal/cnibgp` cannot be vetted whole on this machine: it reaches `internal/plumbing/ebpf`, whose bpf2go artifacts are not committed and cannot be regenerated on macOS (no `llvm-strip`, no `linux/bpf.h`). I type-checked `internal/cnibgp/config.go` against its package types in isolation instead — clean. CI covers the rest.
- Unit tests do not run natively on macOS (linux-only netlink), so CI is the gate there.

privateip (approved):

Issue #327, opened 2026-08-10
A pod can attach with no addresses and nothing says so


ecv:
Addressing config moved inside the `ipam` block, and the master plugins now decide whether to allocate at all purely from whether that block is present. A config written against the old shape still parses cleanly — the moved keys are simply unknown fields, and unknown fields are dropped.

The pod then attaches with a working interface and no addresses, and its BGPAdvertisement is created advertising nothing. Nothing errors, nothing warns. The workload comes up and has no connectivity.

Anything still emitting the old shape — the companion operator being the one that matters — hits this the moment the new plugin lands.

## Outcome

A config carrying the old top-level addressing keys fails loudly at ADD, naming the key and where it moved to. Guessing wrong about a pod's addressing is worse than refusing to attach it.

Related to #304

ecv:
Path, on #304's head:

1. `internal/cni/types.go` — `PluginConf` no longer has `IPv6Subnet`/`IPv4Subnet`/`AddressFamilies`; `encoding/json` drops them silently.
2. `internal/cni/result.go` `buildVethResult` — `if pluginConf.IPAM != nil` is the only gate, so no delegation happens.
3. `internal/cnibgp/bgp.go` `ConfigureHostGateway` — returns nil on a nil result, so no gateway, no route.
4. `internal/cnibgp/bgp.go` `ipamAdvertisementPrefixes` — nil result yields no prefixes, and the BGPAdvertisement is still created.

Same path in `internal/cnitap/ops_add.go`.

The guard is a few lines in each master plugin's `parseConf`: unmarshal into a struct carrying the four moved keys and return a `types.Error{Code: 7}` naming any that are set.

ecv:
Closed by #340 (merged `45e8147`): both master plugins now run `hostconf.RejectMovedIPAMKeys` in `parseConf`, so a config carrying the old top-level addressing keys fails ADD with a code 7 naming them instead of attaching a pod with no addresses.

PR #340, opened 2026-08-10
fix(cni): reject addressing keys left outside the ipam block

ecv:
A pod could come up with a working interface, no addresses at all, and a route advertisement carrying nothing — and nothing anywhere reported a problem.

That happened whenever a network config was still written against the older layout, with the addressing settings sitting at the top level rather than inside the addressing block they moved into; the stale settings were quietly discarded as unrecognised, so the plugin concluded there was no addressing to do.

Attaching now fails outright on such a config, with an error naming each misplaced setting and stating where it belongs, so the operator emitting the old shape gets told exactly what to change.

Guessing wrong about a workload's addressing is worse than refusing to attach it.

## Test plan

- New table-driven cases on both master plugins: one per misplaced addressing setting, one with all four present at once, and one proving the current layout still parses and attaches.
- New cases on the shared guard itself for the edge behaviour: a wrong-typed value is still reported as misplaced, an explicit null reads as absent, and malformed input is left to the caller's own parse error.
- Lint and format checks pass locally over both plugins, cross-compiled for Linux. The plugins themselves are Linux-only and their tests do not run on macOS, so the suites run in CI; the shared guard's own cases were executed locally against an isolated copy and all pass.

Related to #327

### DIFF
--- a/internal/cni/cni_test.go
+++ b/internal/cni/cni_test.go
@@ -62,6 +62,17 @@ func assertCNIError(t *testing.T, err error, wantCode uint, wantMsg string) {
 	}
 }
 
+// movedKeyConf builds an otherwise-valid galactic-cni config carrying the
+// supplied raw JSON member(s) at the top level, for the addressing keys
+// that belong inside the "ipam" block.
+func movedKeyConf(members string) string {
+	return fmt.Sprintf(
+		`{"cniVersion":"1.0.0","name":"test","type":"galactic-cni",`+
+			`"vpc":"%s","vpcattachment":"%s",%s}`,
+		testVPC, testAttachment, members,
+	)
+}
+
 // ---- parseConf -----------------------------------------------------------
 
 func TestParseConf(t *testing.T) {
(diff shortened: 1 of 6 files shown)


ecv:
Implementation detail, for review.

The guard lives in `internal/cni/hostconf` as `RejectMovedIPAMKeys`, next to the conflist loader both master plugins already share. Two copies of the same four-key list would drift; one cannot.

It decodes the stdin bytes into a struct with exactly `ipv6_subnet`, `ipv4_subnet`, `address_families` and `static_ip`, each a `json.RawMessage`. Only presence matters, and `RawMessage` never fails to decode — so `"ipv6_subnet": 42` is still reported as misplaced rather than blowing up the decode. An explicit `null` reads as absent.

A decode error returns nil. Both `parseConf` functions unmarshal the same bytes into their full config first, so malformed JSON has already been rejected with its own message; reporting it twice would only bury the real one.

Both call sites sit right after the base62 identifier checks and before `loadHostConf`, so a stale config is refused before the plugin touches the environment or opens its log file.

The error is `&types.Error{Code: 7}`, matching the identifier-validation errors alongside it. It names every offending key, not just the first:

```
addressing fields 'ipv6_subnet', 'ipv4_subnet' belong inside the 'ipam' block, not at the top level of the config
```

Nothing in `internal/cniipam` changed. Value validation — CIDR shape, family, prefix length — still lives there and still runs on the delegated bytes.

One thing left alone: the existing comment in each `parseConf` said validating these fields here would be redundant. That is still true of their *values*; it was never true of their placement. Both comments now say so and point at the new call.

privateip (approved):

Issue #17, opened 2026-04-08
Galactic VPC - Phase 1 Scope


scotwells:
This issue outlines the work required to integrate Galactic VPC into the Datum Cloud platform and bring it to consumers.

### Dependencies (In Progress)
- [ ] Machine Accounts - Non-human identity for connectors, gateways, and service components to authenticate to the platform
- [ ] IPAM Platform Capability - Address pool allocation system
- [ ] BGP Infrastructure - GoBGP deployment, route reflector topology

### Control Plane & Routing

- Migrate from MQTT to BGP VPNv6 - Replace galactic-router MQTT-based distribution with BGP route reflectors for scalable, standards-based route distribution
- Deploy route reflector topology - HA route reflector cluster with peering to galactic-agents on each node
- Implement BGP-based VPN signaling - Use MP-BGP for VPNv4/VPNv6 route exchange with RT/RD for VPC isolation
- Network services integration - Connect to existing network services operator for programming VPC networks

### Ingress & Gateways

- Envoy Gateway VPC attachment - Attach Envoy Gateway pods to VPC networks for HTTP ingress from internet
- MASQUE gateway integration - Deploy MASQUE gateways as VPC ingress points for connector tunnels
- Gateway API resource mapping - Map HTTPRoute/TCPRoute to VPC backend services

### Egress & NAT

- Egress gateway design - Architecture for outbound internet connectivity from private VPC networks
- NAT gateway implementation - Source NAT for private VPC traffic with IPAM-allocated public IPs
- Egress policy controls - Define what destinations VPC workloads can reach

### Connectors & Client Access

- Iroh + MASQUE client update - Update desktop app to use Iroh transport with MASQUE tunneling
- Headless datum-connect - Server-deployable connector for private network integration
- Connector authentication via Machine Accounts - Connectors authenticate to platform using machine identity
- ConnectorAdvertisement registration - Register reachable networks through each connector instance
- ConnectorAttachment for VPCs - Associate connectors with specific VPC networks

### Security & Policy

- Network Policy enforcement - VPC-level traffic control via galactic-agent (eBPF)
- Security Groups - Stateful firewall rules per VPC attachment
- IAM integration - Role-based access for VPC/attachment/route operations
- Audit logging - Activity records for all VPC mutations (depends on Activity system)

### IPAM Integration

- VPC CIDR allocation - Allocate VPC network ranges from IPAM AddressPools
[...]

PR #349, opened 2026-08-12
feat(gateway): Add edge XDP NAT+LB eBPF datapath

privateip:
## Summary

The edge gateway needs a kernel-level datapath that DNATs a public VIP:port to a backend's SRv6 uSID and forwards it into the fabric, without a userspace hop per packet. This adds that datapath as a self-contained package tree with no Kubernetes or CRD dependency: the XDP program itself, load/attach/pin logic, a Go-side rule table wrapper, kernel feature preflight checks, and a Prometheus collector over its drop-reason counters. First branch in an 8-part stack splitting the edge NAT+LB gateway feature into reviewable, independently buildable PRs — later branches (control-plane engine, CRD reconcilers, the new binary, manifests, lab canary, docs) build on this one.

## Test plan

- [ ] Unit tests pass for the new eBPF plumbing packages (`task test:unit`)

Related to #17

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,10 @@ internal/plumbing/ebpf/prog/usid_bpfel.go
 internal/plumbing/ebpf/prog/usid_bpfel.o
 internal/plumbing/ebpf/prog/usid_bpfeb.go
 internal/plumbing/ebpf/prog/usid_bpfeb.o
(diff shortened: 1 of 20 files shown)


ecv (approved):
comment volume much less offensive here somehow

PR #350, opened 2026-08-12
feat(gateway): Add gateway control-plane engine

privateip:
## Summary

The eBPF datapath from the previous PR needs a control-plane layer that decides what belongs in the rule table and keeps it converged as backends and quotas change. This adds that engine: desired-vs-active rule diffing, a kernel-backed datapath implementation, primary/secondary local-pref and placement logic, per-VPC quota enforcement, and telemetry. It defines its own types rather than reusing the CRD API, so this package has no Kubernetes dependency at all. Second branch in the edge-gateway stack; builds on the eBPF datapath PR.

## Test plan

- [ ] Unit tests pass for the engine package (`task test:unit`)

Related to #17

### DIFF
--- a/internal/gateway/datapath.go
+++ b/internal/gateway/datapath.go
@@ -1,0 +1,138 @@
+// Copyright 2026 Datum Cloud, Inc.
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+package gateway
+
+import "context"
+
+// Datapath is the engine's interface onto the actual NAT/load-balancing
+// dataplane for one DesiredRule. GoBGP, via galactic-router's existing
+// embedding, owns all BGP speaking (the design plan's Active-Active BGP
+// model); no BGP speaker lives in this interface's implementations at
+// all.
+//
+// KernelDatapath (kerneldatapath.go) is the real implementation, backed by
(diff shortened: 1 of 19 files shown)


ecv (dismissed):
these comments should all be struck, and replaced with a design document and verbose commit messages. retrain ur claudes

ecv (approved):
Approving. This is the only branch in the stack with green CI, and the engine reads clean: the mutex discipline is consistent, `Reconcile` converges rather than diffing field by field, and the crash-recovery generation ordering is right.

Four things found, all on failure paths, none blocking. Filed as #359 with the line references.

The one that matters: a rule that passes the quota check and then fails to program holds its reservation forever, because it never enters the active set and so never reaches the release path. `cmd/galactic-gateway` wires the real enforcer, not the noop, so the counters drift on a node that still has room.

The other three are smaller. Teardown skips the release on a datapath failure, which self-heals on retry. `Stop` deletes a rule from the active set on every teardown failure after the first. And a partial register leaves untracked entries in the rule table, which orphan recovery does eventually sweep.

The tests cover the happy paths and quota denial, but nothing asserts quota or active-set state after a failure, which is why none of these show up.

Separately, the comment volume here is tracked in #357 rather than as review comments.

PR #351, opened 2026-08-12
feat(gateway): Add NetworkGateway/NetworkRule controllers and webhook

privateip:
## Summary

The gateway engine from the previous PR needs something to feed it desired state from the cluster. This adds the NetworkGateway and NetworkRule reconcilers that watch those CRDs, resolve each backend's SRv6 uSID, and drive the engine, plus a NetworkRule admission webhook that authorizes VPC ownership before a rule is created. The webhook is wired into galactic-router behind an opt-in flag, disabled by default. Third branch in the edge-gateway stack; builds on the control-plane engine PR.

## Test plan

- [ ] Unit tests pass for the controllers and webhook (`task test:unit`)

> [!NOTE]
> The webhook's ValidatingWebhookConfiguration, Service, and cert-manager wiring don't exist yet. It stays disabled by default, so nothing here is affected, but the feature isn't deployable end-to-end until that follow-up lands.

Related to #17


ecv (commented):
Holding off on approval for one line, not for the controllers.

`go.mod:117` on this branch adds `replace go.datum.net/network => ../network`, and that takes the whole module down anywhere the sibling checkout does not exist. Lint, Build, Unit Tests and Unit Tests (root) are all failing here on `internal/cnibgp/bgp.go`, which has nothing to do with this change. Every branch below inherits it, which is why #350 is the only green PR in the stack. Filed as #358 with the full trail.

The good news is it looks unblocked: `datum-cloud/network`'s default branch already carries `api/v1alpha1/gateway_types.go` and `rule_types.go`, so a re-pin and deleting the replace should be the whole fix, with a compile to confirm the shape has not drifted since your local copy.

On the controllers themselves, no objection. The finalizer-guarded teardown ordering in `NetworkRuleReconciler` is the part I looked hardest at and it holds up: the BGP route is withdrawn before the datapath state goes, which is the order that matters.

One thing I want to flag as intentional rather than as a finding, since a reader landing here cold will wonder: `NetworkGatewayReconciler` and `NetworkRuleReconciler` are not registered with any manager on this branch, and the RBAC for the new resources is not here either. Both arrive later in the stack, in #352 and #354. That is the stack working as designed, not dead code.

Happy to approve as soon as the replace is gone.

ecv (dismissed):

ecv:
Went back through the controllers properly, rather than stopping at the build failure. Four things, filed rather than left inline.

## The one I would fix before this merges

#364. Every gateway node's process watches every gateway object in the namespace, and the node check that scopes work to "my own gateway" reads the target off the object. On a delete there is no object to read, so the not-found branch stops the engine unconditionally. Delete one node's gateway and every other gateway node in that namespace empties its rule table.

That is the opposite of what active-active is for, and it is quiet: each node believes it did the right thing. `TestNetworkGatewayReconciler_StopsEngineOnNotFound` asserts the current behavior with one gateway in the fixture, so a second gateway in that test is probably the whole regression guard.

## Follow-ups

#365. Advertisement wiring, self-address publication and the orphan sweep all log their errors and continue, and the Ready condition is computed from the engine result alone. A node that programmed every rule and advertised none reports `Ready=True`, and nothing retries. The engine failure path right above it does propagate and does set a reason on the condition, which is the shape the other three want.

#367. Two timing gaps. A rule created before any gateway node registers waits for the informer resync, not for the nodes, because this reconciler watches only rules. The comment on that path says a gateway event re-triggers it and then correctly says the opposite in the same sentence. Separately, teardown rebuilds advertisement names from currently-registered nodes, so an advertisement created for a node that has since left is never withdrawn while the finalizer comes off anyway.

#366. The authorization path is complete in code and inert in practice: the placeholder authorizer is what the production caller passes, and no webhook configuration is deployed anywhere in the stack. Worth separating because the Accepted condition is granted on "gateway nodes exist" and is what gates programming, so the name promises a check that never ran. All of it is documented in comments, so nothing here was hidden. It is the next reader I care about.

## What reads well

The teardown ordering is right: route withdrawal precedes NAT release, and the finalizer's doc comment is honest about what it does not coordinate across nodes.

`AssignPrimaryNode` being a pure function of inputs every node observes identically is the right call for a codebase with no leader election, and the "assign exactly once" guard is in the correct place.

Deliberately not reporting: the reconcilers not being registered with any manager here, and the RBAC for the new resources being absent. Both arrive in #352 and #354. That is the stack working.

Still holding approval on #358 rather than on any of the above.

privateip:
fixes #364

ecv (approved):

PR #352, opened 2026-08-12
feat(gateway): Add galactic-gateway binary

privateip:
## Summary

The gateway engine and reconcilers from the previous PRs need a process to run in. This adds galactic-gateway as its own binary, separate from galactic-router's tenant BGP process, so a crash in the XDP-holding gateway engine no longer takes BGP down with it and vice versa. Tenant BGP keeps running in galactic-router, co-located on the same gateway node. Fourth branch in the edge-gateway stack; builds on the controllers and webhook PR.

## Test plan

- [ ] Unit tests pass for the new binary and its config (`task test:unit`)
- [ ] `task build` produces `bin/galactic-gateway` and `task lint` is clean

Related to #17

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -140,6 +140,7 @@ tasks:
       - go build -ldflags "{{.LDFLAGS}}" -o bin/galactic-bgp ./cmd/galactic-bgp
       - go build -ldflags "{{.LDFLAGS}}" -o bin/galactic-route ./cmd/galactic-route
       - go build -ldflags "{{.LDFLAGS}}" -o bin/galactic-router ./cmd/galactic-router
+      - go build -ldflags "{{.LDFLAGS}}" -o bin/galactic-gateway ./cmd/galactic-gateway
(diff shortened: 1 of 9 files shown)


ecv (dismissed):
Approving on content. CI here is red from the module replace inherited from #351, tracked in #358, not from anything in this PR.

The binary follows the `galactic-router` precedent closely, and the two things I most wanted to check are right: no leader election, which is correct for a per-node DaemonSet where every node needs its own datapath, and complete cleanup on every error path in `setupGatewayDatapath`.

The keep-alive var deserves the comment it has. A GC'd link silently detaching while every control-plane signal still reads healthy is a genuinely nasty failure, and it is worth having found that live rather than in production.

Four follow-ups filed as #360. The one worth acting on is ordering: the health server answers `SERVING` before the datapath is attached, so there is a window where the probe says ready and no traffic can be intercepted. Bounded, since a failed attach exits the process, and there is no Service routing on readiness. Still the wrong way round.

The rest are small: a health server that stops serving is logged past rather than fatal, an IPv4 gateway address passes startup validation and fails deeper in, and the interface lookup sits after the attach that would already have failed on it.

ecv (approved):

PR #353, opened 2026-08-12
feat(gateway): Publish galactic-gateway as its own image

privateip:
## Summary

The new binary from the previous PR needs a release image and a way to get it published. This adds a Dockerfile for galactic-gateway, matching galactic-router's shape, and a publish job that pushes it and stamps its tag into the two-container gateway DaemonSet's OCI Kustomize bundle entry. Fifth branch in the edge-gateway stack; builds on the galactic-gateway binary PR.

## Test plan

- [ ] `docker build` succeeds for the new Dockerfile
- [ ] The publish workflow YAML is valid and `task lint` is clean

Related to #17

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -32,6 +32,19 @@ jobs:
       platforms: linux/amd64,linux/arm64
     secrets: inherit
 
+  publish-galactic-gateway-image:
+    permissions:
+      id-token: write
+      contents: read
+      packages: write
+      attestations: write
+    uses: datum-cloud/actions/.github/workflows/publish-docker.yaml@v1.20.0
+    with:
+      image-name: galactic-gateway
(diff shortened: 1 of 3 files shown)


ecv (dismissed):
Approving. Small and correct: the new image job is wired the same way as the others, `publish-kustomize-bundles` correctly `needs` all four, and stamping `config/gateway/base` twice for the two containers in that pod is right.

CI red here is inherited from #351's module replace, tracked in #358.

One thing to carry into that fix rather than this PR. `publish.yaml` passes no `build-contexts` input to the reusable publish workflow, and #355 later adds `COPY --from=network . /network` to both the router and gateway Dockerfiles. Once that lands, these publish jobs cannot build, including the router image that ships today. Both Dockerfile lines need to come out with the replace.

scotwells (approved):

PR #354, opened 2026-08-12
feat(gateway): Add config/gateway/ Kustomize manifests

privateip:
## Summary

The image published in the previous PR needs Kubernetes manifests to actually run it. This adds the gateway control plane's own ServiceAccount, ClusterRole, and two-container DaemonSet base, split out of galactic-router's manifests now that the reconcilers they cover no longer live in that binary. galactic-router's RBAC is trimmed to match, and the fabric underlay DaemonSet's affinity is extended to gateway-role nodes. Sixth branch in the edge-gateway stack; builds on the images and CI PR.

## Test plan

- [ ] `kubectl kustomize` builds cleanly for the new and changed manifests
- [ ] `task lint` is clean

Related to #17


ecv (dismissed):
Approving. Nothing found here, and I went looking specifically at the security surface since this is the manifest that grants a pod the right to attach an XDP program.

The capability set is scoped rather than blanket: `NET_ADMIN`, `BPF` and `PERFMON` on the gateway container, `NET_ADMIN` alone on the co-located router, `allowPrivilegeEscalation: false` on both, and no `privileged: true` anywhere. That is the right shape.

RBAC matches what the reconcilers actually do, including omitting create and delete on the resources the external operator owns. The `bpf-fs` volume being `Directory` rather than `DirectoryOrCreate` is the correct call for a host mount that must already exist.

Keeping this out of the root kustomization is right too, for the same reason `config/fabric` is: the DaemonSet needs a per-node overlay before it means anything, and a generic default would deploy something broken.

CI red here is inherited from #351's module replace, tracked in #358.

ecv (approved):

scotwells (approved):

PR #378, opened 2026-08-13
docs: Align egress gateway NAT plan with enhancements#865

privateip:
## Summary

The edge gateway's egress NAT plan wasn't tracked against a real issue. The tracking issue has since been rewritten from static NPTv6 prefix translation to stateful NAT66 (ULA source rewritten to a public GUA address, session-tracked), which is the design this plan already covered under a different name. This update aligns the plan with the current issue text and closes a correctness gap the rewrite surfaced: tenant ULA space isn't guaranteed globally unique, so sessions now key on a per-tenant discriminator rather than the tenant's own address alone.

## Test plan

- [ ] Plan reads consistently end to end and cross-links to the issue it now matches

Related to datum-cloud/enhancements#865


aflor024 (approved):
Approved

PR #382, opened 2026-08-13
docs: Add gateway packet-trace and C4 architecture diagrams

privateip:
## Summary

The gateway packet-trace diagram described the datapath incorrectly — plain DNAT instead of Full-NAT, the backend's reply routed to the client instead of the gateway, and worker-side decap attributed to a kernel SRv6 feature this codebase doesn't use — and the architecture diagrams didn't call out the edge gateway as its own component. This validates and corrects the packet-trace diagrams against the actual datapath source, and adds a System Context callout plus a new component-level diagram for the gateway's internals.

## Test plan

- [ ] Diagrams render cleanly on GitHub (no Mermaid parse errors)
- [ ] Packet-trace and component diagrams match the current datapath/reconciler behavior


docs/architecture/README.md
scotwells, line 1: I wouldn't worry about Layer 3 at all with C4 diagrams. Those typically get out of sync quick. I think the Context digram you have is actually a better Container diagram since it shows the individual components that make up the galactic system. The Context digram should primarily stick with product-level view on why something exists (e.g. center of all networking functionality for Connectors, Compute, Gateways, etc).

docs/gateway/gateway-ingress-packet-trace.md
scotwells, line 1: Probably don't need file paths in here?

scotwells (dismissed):
Just some minor feedback, nothing holding back approval.

scotwells (approved):

PR #385, opened 2026-08-13
feat(gateway): reconcile tenant VRF egress default routes

### LINKED ISSUE #384: Decide default-route ownership for egress (masquerade) — #865 Phase D
## Context

Phase D of [#865](https://github.com/datum-cloud/enhancements/issues/865) (NAT masquerade egress for the edge gateway), per [`docs/plans/865-edge-gateway-nat66-egress.md`](https://github.com/datum-cloud/galactic/blob/main/docs/plans/865-edge-gateway-nat66-egress.md) §4.4/§7.1. Phases A-C (#380, #381, #383) are buildable and testable in isolation without this decision. No egress packet reaches a gateway node in a real cluster until it's resolved — this is the plan's largest open question.

## The question

Something needs to install a tenant VRF's `::/0` (or public-internet-covering aggregate) route encapsulating toward that tenant's assigned `egress_sid`, via the same `internal/plumbing/srv6.RouteEgressAdd` every other tenant-VRF SRv6 route already uses. Nothing in this repo triggers that today.

## Candidate 1: `galactic-cni`'s `cmdAdd` installs the route at pod-attach time

- Fits the existing "attach-time sets up everything" model the CNI already uses for every other route.
- Requires the CNI to gain a live Kubernetes API read (`NetworkEgressPolicy`, `NetworkGateway`) it doesn't have today — CNI currently only acts on identifiers pre-populated into its config, never calls out to the API at attach time.
- One-shot: a tenant enabling egress after pods are already running never gets the route until pods restart. Disabling egress doesn't revoke it until restart either. Gateway-node reassignment (if `AssignPrimaryNode`'s target shifts) has the same staleness problem.
- The VRF is shared across every attachment on a VPC, not per-pod, so `cmdAdd`/`cmdDel` would need "am I first/last attachment" bookkeeping they don't have today.

## Candidate 2: a new per-node controller

Watches `NetworkEgressPolicy` + `VPCAttachment`, reconciles the route against every tenant VRF present on its node.

- Self-heals: enablement, disablement, and gateway-node reassignment are all picked up on the next reconcile pass, no pod restart needed.
- Keeps `galactic-cni` exactly as architecturally simple as it is today.
- Mirrors an idiom this repo already trusts (`gc_controller.go`'s node-scoped kernel-state reconciliation) rather than introducing a new one.
- Trade-off: a brand-new pod has a short window with no route until this controller's next pass notices the new VRF — the same class of lag every other reconciler in this repo already has.

## Recommendation

Candidate 2. Candidate 1's retroactive enable/disable gap isn't a style preference — it's a correctness bug ("I turned off egress and traffic kept flowing").

## Resolved already

Node-target resolution (which gateway node's `egress_sid` a tenant's route should point at): reuse `AssignPrimaryNode(vpcRef)` unmodified, per §4.5 — same node as that tenant's primary ingress node. Confirmed, not blocked on this issue.

## Next step

Decide the route-owner question above, then resume Phase D implementation.

### LINKED ISSUE #389: Design: tenant VRF egress default-route reconciler (865 Phase D)
Part of datum-cloud/enhancements#865. Phase D of `docs/plans/865-edge-gateway-nat66-egress.md`. Resolves #384. Implemented in #385.

## Purpose

Decide and implement who installs a tenant VRF's `::/0` route toward its assigned gateway's `egress_sid` — the plan's own largest open question (§4.4/§7.1): nothing in the datapath reaches a gateway node in a real cluster until this exists.

## Decision (#384)

Two candidates: (1) `galactic-cni`'s `cmdAdd` installs the route at pod-attach time, or (2) a new per-node controller reconciling continuously. Chose (2): candidate (1) requires the CNI to gain a live Kubernetes API dependency it doesn't have today, and — more importantly — is one-shot: a tenant enabling or disabling egress after pods already exist would never see the change take effect without a pod restart, a correctness gap, not a style preference. Candidate (2) self-heals on enable/disable/gateway-node reassignment and mirrors an idiom this repo already trusts (`gc_controller.go`'s node-scoped kernel-state reconciliation).

### `NetworkEgressPolicyReconciler` (new, runs in `galactic-gateway`)

One-time `AssignedGatewayNode` assignment, mirroring `NetworkRuleReconciler.assignPrimaryNode` almost exactly (same immutability guard, same Accepted-condition-since-webhook-can't-set-it reasoning). No finalizer: `NetworkEgressPolicy` owns no `BGPAdvertisement` to withdraw in a particular order; deletion just means the route reconciler stops seeing it as accepted on its next pass.

### `internal/egressroute` + `EgressRouteReconciler` (new, runs in `galactic-router`, every compute node)

Ticker-driven (mirrors `GCReconciler`'s split of a real-logic package plus a thin controller wrapper). Per reconcile pass: enumerate local Galactic-managed VRF interfaces (`vrf.ResolveVPC`, new), and for each VPC:

1. Is there an accepted `NetworkEgressPolicy` for it, with a resolved `AssignedGatewayNode`?
2. Does that `NetworkGateway` publish a non-empty `EgressSID`?
3. Does this node have its own `BGPVRFInstance` for the VPC?

If all three: compose the destination uSID (that gateway's `EgressSID` locator + this VPC's own `tenant_arg`, see below) and `RouteEgressAdd` the `::/0` route. Otherwise, remove it if present.

Runs in `galactic-router`, not `galactic-gateway`, because tenant VRFs live on compute nodes, not gateway nodes — `galactic-router`'s tenant role already runs everywhere one could exist.

## Two inferences beyond the plan's text

1. **`tenant_arg` reuses each VPC's own `BGPVRFInstance.Spec.VRFID`** — already allocated per-node for ingress SRv6 decap (`internal/cnibgp/bgp.go`'s `allocateArgument`), and already unique *per-node* (not globally per-VPC — allocation scans only that node's own instances), which is exactly the isolation property egress needs: no two VPCs on the same node share a value. Two different compute nodes hosting the same VPC can pick different values without issue, since nothing downstream ever compares one node's choice against another's for the same VPC.
2. **`NetworkGatewayStatus.EgressSID`** (network#16) had to be published into BGP the same way `SRv6Address` already is — `RouteEgressAdd`'s underlying `netlink.RouteGet` requires a real kernel route to the destination to already exist, which only a BGP/EVPN advertisement provides.

Neither is spelled out in the plan's own text; both were necessary to make this phase actually work.

## Bug found and fixed along the way

`srv6.RouteEgressDel` has been broken since it was written — it set an empty `SEG6Encap{}` on the delete request, which `netlink`'s `EncodeSEG6Encap` unconditionally rejects (`"No Segment in srh"`). Already called in production by `internal/runtime/gobgp/monitor.go`'s BGP path-withdrawal handler, so every SEG6 route withdrawal in this codebase was likely failing, entirely unrelated to egress. Fixed, with the first test coverage `RouteEgressDel` has ever had.

## References

- Design plan: `docs/plans/865-edge-gateway-nat66-egress.md` §4.4-4.5, §7.1
- Design discussion and decision: #384
- Implemented in #385 (stacked on #383)

privateip:
## Summary

Phase D of #865, per docs/plans/865-edge-gateway-nat66-egress.md
§4.4/§4.5. Resolves #384 (candidate 2: new controller). Stacked on
#383 (Phase C). Depends on datum-cloud/network#15 (now includes the
Phase D CRD field additions too).

- `internal/egressroute`: per-node reconcile logic -- enumerate local
  VRFs, resolve VPC, check `NetworkEgressPolicy` enablement + gateway
  assignment, resolve the destination uSID, install/remove the `::/0`
  SEG6 encap route. Mirrors `internal/gc`'s split of real-logic-package
  plus thin-controller-wrapper.
- `internal/controller/egressroute_controller.go`: ticker-driven
  wrapper (mirrors `GCReconciler`), registered in `galactic-router` --
  runs on every compute node, not just gateway nodes.
- `internal/controller/networkegresspolicy_controller.go`: new
  `NetworkEgressPolicyReconciler`, one-time gateway-node assignment
  mirroring `NetworkRuleReconciler.assignPrimaryNode`. Registered in
  `galactic-gateway`.
- `internal/plumbing/vrf.ResolveVPC`: new exported VPC-from-VRF-name
  resolver.
- `NetworkGatewayReconciler.publishEgressAddresses`: extended to
  also publish `EgressSID`; refactored to share a new
  `publishHostRouteAdvertisement` helper across all three
  self-address advertisements instead of triplicating the
  get-or-create-or-update block.
- Config: `GALACTIC_ROUTER_EGRESS_ROUTE_INTERVAL`, separate from GC's
  own interval.

## Two inferences beyond the plan's text

Neither is spelled out in the design plan; both were necessary to
make this phase actually work (full reasoning in
`internal/egressroute`'s package doc comment):

- `tenant_arg` reuses each VPC's own `BGPVRFInstance.Spec.VRFID`
  (already allocated per-node for ingress SRv6 decap, and already
  unique per-node -- exactly the isolation property egress needs).
- `NetworkGatewayStatus.EgressSID` must be published into BGP the
  same way `SRv6Address` is, or no compute node has a kernel route to
  encapsulate toward at all.

## Bug found and fixed along the way

`srv6.RouteEgressDel` has been broken since it was written -- it set
an empty `SEG6Encap{}` on the delete request, which netlink
unconditionally rejects (`"No Segment in srh"`). Already called in
production by `internal/runtime/gobgp/monitor.go`'s BGP
path-withdrawal handler, so every SEG6 route withdrawal in this
codebase was likely failing, unrelated to egress. Fixed, with the
first test coverage `RouteEgressDel` has ever had.

## Testing

- New tests: `internal/egressroute` (unit + root-gated e2e proving
  the full install/remove lifecycle against a real kernel VRF),
  `internal/plumbing/vrf`, `internal/plumbing/srv6` (the bug-fix
  regression test), `internal/controller` (both new reconcilers),
  `internal/config`.
- `go build ./...`, `go vet ./...`, `task lint` (0 issues) -- clean.
- Full `task test:unit` under root: everything this PR touches
  passes. The only failure present (`internal/cni/tap`) is the same
  pre-existing, unrelated environment issue confirmed on earlier PRs
  in this stack.

Closes #384.

### DIFF
--- a/cmd/galactic-router/root.go
+++ b/cmd/galactic-router/root.go
@@ -282,6 +282,14 @@ func runCmd(cfg *config.RouterConfig) error {
 		}
 	}()
 
+	// Register EgressRoute controller and start its ticker goroutine
+	// (datum-cloud/enhancements#865, design plan §4.4/§7.1) -- extracted
+	// to its own function to keep this function's own cyclomatic
+	// complexity down (gocyclo), not for any reuse reason.
+	if err := startEgressRouteController(ctx, mgr, cfg, nodeName); err != nil {
+		return err
+	}
+
 	if err := mgr.Start(ctx); err != nil {
 		return fmt.Errorf("manager exited: %w", err)
 	}
@@ -292,6 +300,52 @@ func runCmd(cfg *config.RouterConfig) error {
 	if cause := context.Cause(ctx); cause != nil && !errors.Is(cause, context.Canceled) {
 		return cause
 	}
+
+	return nil
+}
+
+// startEgressRouteController registers controller.EgressRouteReconciler and
+// starts its ticker goroutine, reconciling tenant VRF ::/0 default routes
+// toward each VPC's assigned gateway egress_sid (datum-cloud/
+// enhancements#865, design plan §4.4/§7.1). Mirrors the GC ticker in
+// runCmd exactly (including waiting for cache sync first, so the initial
+// pass doesn't see an empty NetworkEgressPolicy list and remove every live
+// egress route) -- reuses cfg.GCNamespace (both scan the same namespace's
+// CRDs) but its own separate interval, see DefaultRouterEgressRouteInterval's
+// doc comment for why.
+func startEgressRouteController(
+	ctx context.Context, mgr ctrl.Manager, cfg *config.RouterConfig, nodeName string,
+) error {
+	rec := &controller.EgressRouteReconciler{
+		Client:    mgr.GetClient(),
+		Scheme:    mgr.GetScheme(),
+		Namespace: cfg.GCNamespace,
+		NodeName:  nodeName,
+		Interval:  cfg.EgressRouteInterval,
+	}
+	if err := rec.SetupWithManager(mgr); err != nil {
+		return fmt.Errorf("setup EgressRoute controller: %w", err)
+	}
+
+	go func() {
+		ticker := time.NewTicker(cfg.EgressRouteInterval)
+		defer ticker.Stop()
+
+		if !mgr.GetCache().WaitForCacheSync(ctx) {
+			log.Printf("EgressRoute: cache sync failed, skipping initial pass")
+			return
+		}
+		rec.RunOnce(ctx)
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				rec.RunOnce(ctx)
+			}
+		}
+	}()
 
 	return nil
 }
@@ -344,6 +398,9 @@ func newRootCommand() *cobra.Command {
 	cmd.Flags().DurationP("gc-interval", "",
 		config.DefaultRouterGCInterval,
 		"Cleanup interval")
+	cmd.Flags().DurationP("egress-route-interval", "",
+		config.DefaultRouterEgressRouteInterval,
+		"Egress default-route reconcile interval (datum-cloud/enhancements#865)")
 	cmd.Flags().Bool("webhook-enabled", false,
 		"Enable the NetworkRule admission webhook (requires TLS cert material; see config/webhook/)")
 	cmd.Flags().IntP("webhook-port", "",
--- a/cmd/galactic-gateway/root.go
+++ b/cmd/galactic-gateway/root.go
@@ -138,6 +138,7 @@ func runCmd(cfg *config.GatewayConfig) error {
 		NodeName:      nodeName,
 		SRv6Address:   cfg.SRv6Address,
 		EgressAddress: cfg.EgressAddress,
+		EgressSID:     cfg.EgressSID,
 	}).SetupWithManager(mgr); err != nil {
 		return fmt.Errorf("setup NetworkGateway controller: %w", err)
 	}
@@ -151,6 +152,17 @@ func runCmd(cfg *config.GatewayConfig) error {
 		NodeName: nodeName,
 	}).SetupWithManager(mgr); err != nil {
 		return fmt.Errorf("setup NetworkRule controller: %w", err)
+	}
+
+	// Register NetworkEgressPolicy controller (one-time gateway-node
+	// assignment; see internal/controller/networkegresspolicy_controller.go,
+	// datum-cloud/enhancements#865).
+	if err := (&controller.NetworkEgressPolicyReconciler{
+		Client:   mgr.GetClient(),
+		Scheme:   mgr.GetScheme(),
+		NodeName: nodeName,
+	}).SetupWithManager(mgr); err != nil {
+		return fmt.Errorf("setup NetworkEgressPolicy controller: %w", err)
 	}
 
 	if err := mgr.Start(ctx); err != nil {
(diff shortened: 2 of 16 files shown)


cmd/galactic-router/root.go
```diff
@@ -296,6 +304,52 @@ func runCmd(cfg *config.RouterConfig) error {
...
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				rec.RunOnce(ctx)
+			}
+		}
+	}()
```
scotwells, line 348: Why's this use a separate ticker process for reconciling things instead of leveraging the normal controller runtime behavior?

cmd/galactic-router/root.go
```diff
@@ -344,6 +398,9 @@ func newRootCommand() *cobra.Command {
 	cmd.Flags().DurationP("gc-interval", "",
 		config.DefaultRouterGCInterval,
 		"Cleanup interval")
+	cmd.Flags().DurationP("egress-route-interval", "",
+		config.DefaultRouterEgressRouteInterval,
+		"Egress default-route reconcile interval (datum-cloud/enhancements#865)")
```
scotwells, line 403: The link to an enhancement here is a little odd.

PR #411, opened 2026-08-15
fix: Restore IPv6-only Cilium networking in containerlab labs

privateip:
## Summary

CoreDNS (and any pod scheduled off the edge-labeled node) sat at 0/1 Ready in the containerlab labs because Cilium silently fell back to IPv4-only pod networking, which can't reach these clusters' IPv6-only Kubernetes service network. The `--set` flags meant to force Cilium into IPv6-only mode referenced keys that don't exist in any Cilium chart, so they were ignored. This switches to the real chart keys and makes the install step idempotent so a rerun actually applies the fix instead of silently keeping the broken config.

> [!NOTE]
> This fixes the install path for a fresh cluster. Live-upgrading an already-broken cluster to IPv6-only may still need manual repair of stale per-node IPAM state left over from the IPv4 allocation.

## Test plan

- [ ] Fresh `task deploy` brings up Cilium IPv6-only and CoreDNS reaches Ready on all three sites
- [ ] Re-running `deploy-cni.sh` against an already-provisioned cluster upgrades Cilium instead of silently no-op'ing

### DIFF
--- a/deploy/containerlab/scripts/deploy-cni.sh
+++ b/deploy/containerlab/scripts/deploy-cni.sh
@@ -29,24 +29,40 @@ for site in dfw sjc iad; do
   echo "Installing galactic-cni on ${site}..."
 
   # Cilium — installed via the cilium CLI (no Helm binary/repo needed;
-  # the CLI vendors the chart itself). What looked like the CLI
-  # "ignoring --set" was actually stale value keys: Cilium's chart
-  # flattened ipv4.enabled/ipv6.enabled into top-level enableIPv4/
-  # enableIPv6, so the old nested --set was a silent no-op against the
-  # current chart and Cilium fell back to its IPv4 default, allocating
-  # from 10.0.0.0/8 — unreachable from these IPv6-only clusters
-  # (ipFamily: ipv6). The clusters need Cilium told to disable IPv4 via
-  # the current flat keys. Tunnel mode is used since Cilium v1.20
-  # requires IPv4 for native routing mode with cluster-pool IPAM.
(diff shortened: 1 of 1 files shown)


deploy/containerlab/scripts/deploy-cni.sh
```diff
@@ -29,24 +29,40 @@ for site in dfw sjc iad; do
...
-      --set enableIPv6=true \
+    CILIUM_ACTION=install
+    cilium status &>/dev/null && CILIUM_ACTION=upgrade
+    cilium \${CILIUM_ACTION} \
+      --set ipv4.enabled=false \
+      --set ipv6.enabled=true \
       --set ipam.mode=cluster-pool \
-      --set clusterPoolIPv6.clusterCIDR=fd00:100::/48 \
-      --set clusterPoolIPv6.maskSize=64 \
+      --set ipam.operator.clusterPoolIPv6PodCIDRList='{fd00:100::/48}' \
```
0xmc: Perhaps set this as a variable

0xmc (dismissed):
A quasi-nit noted.  Otherwise looks good.

deploy/containerlab/scripts/deploy-cni.sh
privateip: Done in 2efcc3a — pulled the CIDR and mask size into `CILIUM_POD_CIDR_V6`/`CILIUM_POD_CIDR_V6_MASK_SIZE` next to the other version vars, with a note that they need to stay in sync with `podSubnet` in `node_files/{dfw,sjc,iad}/config.yaml`.

ecv (approved):

Issue #331, opened 2026-08-10
An incomplete plugin chain still reports a successful attach


ecv:
The master plugin used to publish BGP and SRv6 state itself, so a successful attach meant the datapath existed. Now that work belongs to a separate plugin the runtime invokes from the conflist, and no one checks it was there.

A conflist missing that entry produces a pod with an interface, addresses, and no path to its VPC. Every plugin that ran returns success.

The chain is authored outside this repo, so a stale or hand-edited conflist is the likely way in.

## Outcome

Attach fails when the chain is incomplete, rather than handing back a pod that looks attached and cannot reach anything.

Related to #305, same failure shape as #327

ecv:
Before #305, `internal/cni/ops_add.go` called `cnibgp.PublishBGPState` in-process. After it, the master's ADD ends at `buildVethResult` and the BGP/SRv6/eBPF work happens only if the conflist names `galactic-bgp`.

`galactic-bgp` does guard its own side — `inferFromPrevResult` errors when `prevResult` is absent, so it cannot run un-chained. Nothing guards the other direction.

Cheapest version: the master checks that a plugin of its own chain's expected type follows it in the conflist it can already read (`internal/cni/hostconf`), and fails ADD if not.

PR #412, opened 2026-08-15
fix(cni): grant galactic-cni get on network-attachment-definitions

privateip:
## Summary

`d3c1ca3` (#331) added a `Get` call on the pod's
`NetworkAttachmentDefinition` before creating any kernel state
(`internal/nadpatch.VerifyChainComplete`), but the `galactic-cni`
ClusterRole only granted `patch` on `network-attachment-definitions` —
the verb the later NAD-annotation step needs, not this new pre-check.

Every ADD failed Forbidden before any interface, VRF, or route was
created, wedging every pod attaching through the "private" NAD chain in
`ContainerCreating`.

## Test plan

- Applied to all three `deploy/containerlab` clusters; `task verify`
  passes (bgp-transit, bgp-fabric, bgp-peers, srv6, evpn, gateway,
  ns10–ns40 scenarios all green).

### DIFF
--- a/config/galactic-cni/rbac.yaml
+++ b/config/galactic-cni/rbac.yaml
@@ -15,7 +15,7 @@ rules:
   - apiGroups: ["k8s.cni.cncf.io"]
     resources:
       - network-attachment-definitions
-    verbs: ["patch"]
+    verbs: ["get", "patch"]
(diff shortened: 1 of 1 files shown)


0xmc (approved):
lgtm

PR #413, opened 2026-08-16
fix: Harden DaemonSet manifests and fix ungraceful BGP shutdown

privateip:
## Summary

A production-readiness review of galactic's DaemonSets against the CIS Benchmark and Pod Security Standards found several gaps in privilege scoping, health/lifecycle probes, and namespace policy visibility. It also surfaced that galactic-router had no graceful shutdown story: it exited on termination without withdrawing its BGP routes, leaving peers to notice only via a timeout instead of an immediate withdrawal. This closes the mechanical hardening gaps, makes every container's actual listening port explicit (catching two collision-prone port defaults along the way), and fixes the shutdown behavior. One cluster-wide secrets-access finding is intentionally left open pending a separate design conversation.

## Test plan

- [ ] DaemonSets schedule and reach Ready with the new security context (verified live on a disposable Kind cluster)
- [ ] galactic-router withdraws BGP routes and shuts down cleanly on SIGTERM within the termination grace period
- [ ] Lint, build, unit tests, and e2e all pass

### DIFF
--- a/cmd/galactic-cni/main.go
+++ b/cmd/galactic-cni/main.go
@@ -68,7 +68,14 @@ func newRunCommand() *cobra.Command {
 		},
 	}
 	runCmd.Flags().IntVar(&grpcHealthPort, "grpc-health-port", 5180, "gRPC health check port")
-	runCmd.Flags().IntVar(&metricsPort, "metrics-port", 9091, "Prometheus metrics HTTP port")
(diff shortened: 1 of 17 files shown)


privateip:
Closing — scoped wrong, opening a replacement with only the router graceful-shutdown commit.

PR #424, opened 2026-08-18
feat: Add ingress sidecar for VPC backend connectivity

privateip:
## Summary

HTTP Ingress for VPC Networks (#796) needs a way to attach the shared Envoy Gateway fleet's node-local Linux VRF/SRv6 route state to each VPC backend pod, so Envoy's own upstream sockets can reach a tenant's pods over the SRv6 underlay. This adds that piece: `galactic-vrf`, a new sidecar binary that runs as a second container in the Envoy Gateway pod and watches the per-pod `EndpointSlice`s galactic-cni publishes (#854).

Two kernel resources are reconciled at two different granularities, matching how the underlying VRF/SID primitives actually work rather than the tenant label they're discovered through: one Linux VRF device per VPC (shared across every attachment of that VPC on the node), and one SRv6 egress route per pod (since pods of the same tenant on different nodes carry different SIDs). Teardown of either is delayed by a configurable grace period after the corresponding EndpointSlice disappears, and a VPC's own grace period never starts while any of its routes are still within their own — so a slow config-push on the extension-server side (#856/#857) can't get raced into a blackhole, and a live sibling attachment is never torn down out from under.

Startup inventories existing kernel state before the first reconcile so a restart doesn't either treat live pods as stale or leak truly orphaned VRFs forever.

Full design in `docs/plans/855-ingress-sidecar-vpc-backend-connectivity.md`.

## Test plan

- [x] Unit tests for the reconciler's desired-state diffing at both granularities (VRF appears/disappears with first/last pod; a route's own lifecycle is independent of its siblings; a second attachment of the same VPC doesn't disturb the first)
- [x] Full-repo build/vet/test clean
- [ ] **Required before merge, not yet done:** real-kernel verification of `RouteEgressAdd`'s `netlink.RouteGet` assumption and the VRF lock's flock-path writability, run from an actual Envoy Gateway pod's netns
- [ ] **Required before merge, not yet done:** end-to-end verification that the existing TC-BPF uSID decap datapath handles this sidecar's traffic the same as EVPN-originated traffic

Opening as draft until both required kernel/e2e checks above are done — see §7 of the plan.

Related to datum-cloud/enhancements#855
Related to datum-cloud/enhancements#796


0xmc (dismissed):
There may be a startup race.  Inventory relies on WaitForCacheSync, which only guarantees the informer's initial list, not that the reconcile queue has drained.

On a busy node at boot this can seed a duplicate boot/<vpc>/<prefix> route entry, and once its grace period expires, Sweep can delete a route a live pod still needs.

Not covered by existing tests (which only test the intended ordering).

Either drain the initial reconcile batch or base Inventory's dedup on live API state.

privateip:
Thanks for catching this @0xmc — confirmed the race and pushed a fix in b383d58.

Root cause: `Store.Inventory` was gated on `mgr.GetCache().WaitForCacheSync(ctx)` alone, which only guarantees the informer's initial `List` landed in the cache — not that the resulting workqueue had been drained by `Reconcile`. On a busy node at boot, `Inventory` could see a still-live pod's kernel route as orphaned before its `Reconcile` caught up, seed it under a synthetic `boot/<vpc>/<prefix>` key with its own grace period, and once that period elapsed, `Sweep` would delete the underlying kernel route (routes are addressed by prefix+table, not by `Store`'s map key) out from under the still-live pod.

Fix: added `ingresssidecar.SeedFromAPI`, which lists every tenant-labeled `EndpointSlice` directly via the manager's uncached reader (`mgr.GetAPIReader()`, safe to use before `mgr.Start`) and applies each one through the same `Store.SetDesired` path `Reconciler` itself uses — before `Inventory` ever runs. This closes the race independently of cache/workqueue timing (your "base Inventory's dedup on live API state" option) rather than trying to drain/wait out the workqueue. `SetDesired` is idempotent, so the controller's own later, now-redundant `Reconcile` of the same objects is harmless.

Also added `internal/ingresssidecar/seed_test.go`, including `TestSeedFromAPIThenInventoryDoesNotOrphanLiveRoute`, which runs the fixed startup order directly and asserts a long sweep afterward never removes a route claimed only through `SeedFromAPI` (no completed `Reconcile` standing in for it) — the gap you noted wasn't covered by the existing tests. Updated `docs/plans/855-...md` §9 item 2 to describe the corrected mechanism.

scotwells (dismissed):

scotwells (approved):

PR #426, opened 2026-08-19
feat: Add Maglev consistent-hash ring and RFC 6296 NPTv6 translation

privateip:
## Summary

The DSR/Maglev gateway redesign needs two algorithms neither the codebase nor its dependencies provide: a Maglev consistent-hash ring for backend selection, and RFC 6296 NPTv6 for stateless per-VRF prefix translation. Both land here as plain Go packages with no eBPF or controller wiring, so their correctness can be reviewed and tested in isolation before anything depends on them.

Building against the new CRD types this needs (`ServiceVIPBinding`, `NAT66Shard`, `BGPVRFInstanceSpec.NPTv6`) required a temporary local `replace` pointing `go.datum.net/network` at a sibling checkout. That resolves on a dev machine but breaks every CI runner (no such directory exists there), so a second commit here pins straight to the published module version instead, skipping the replace entirely.

## Test plan

- [ ] `task test`
- [ ] `task ci` (module resolution succeeds with no sibling `../network` checkout)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,8 @@ require (
 	github.com/spf13/pflag v1.0.10
 	github.com/spf13/viper v1.21.0
 	github.com/vishvananda/netlink v1.3.2-0.20260803231012-156a440d85e5
-	go.datum.net/network v0.0.0-20260819160013-45d0ff9deaee
+	github.com/vishvananda/netns v0.0.5
+	go.datum.net/network v0.0.0-20260814001919-96677d648672
 	golang.org/x/sys v0.47.0
 	golang.org/x/term v0.45.0
 	google.golang.org/grpc v1.83.0
@@ -71,7 +72,6 @@ require (
 	github.com/spf13/afero v1.15.0 // indirect
 	github.com/spf13/cast v1.10.0 // indirect
 	github.com/subosito/gotenv v1.6.0 // indirect
-	github.com/vishvananda/netns v0.0.5 // indirect
 	github.com/x448/float16 v0.8.4 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	go.uber.org/zap v1.28.0 // indirect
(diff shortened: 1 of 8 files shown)


scotwells (dismissed):

privateip:
CI is failing on `Build`/`Lint`/`Unit Tests` with:

```
internal/controller/networkgateway_controller.go:268:18: rule.Status.PrimaryNode undefined
(type v1alpha1.NetworkRuleStatus has no field or method PrimaryNode)
```

This isn't a mirror/infra flake (unlike the apt-get cancellations elsewhere in CI right now) — it's a real break. `go.mod` on this branch pins `go.datum.net/network@v0.0.0-20260819160013-45d0ff9deaee`, which intentionally removed `NetworkRuleStatus.PrimaryNode`. The upstream module even has a regression test documenting why:

> `TestNetworkRuleStatusHasNoPrimaryNode`: the earlier active-passive Full-NAT design assigned a single primaryNode per rule. This design's anycast/DSR model has every NetworkGateway serve every rule identically, so no such field should exist any more.

So `go.mod` was bumped to the post-redesign API, but the galactic-side code here wasn't migrated to match. Three places still implement the old primary/secondary model built around that field:

- `internal/gateway/localpref.go` — `PrimaryLocalPref`/`SecondaryLocalPref` (100 vs 50) keyed on `status.primaryNode`
- `internal/gateway/placement.go` — `AssignPrimaryNode(vpcRef, gatewayNodes)`, a hash-based single-node assignment
- `internal/controller/networkgateway_controller.go:268,275` — reads `rule.Status.PrimaryNode` to skip unassigned rules and compute local-pref via `gateway.LocalPreference(nodeName, primaryNode)`

None of that has an equivalent in the new anycast/DSR model — there's no "primary" left to assign or key local-pref off of, so this needs the gateway local-pref/placement logic replaced, not patched around a missing field.

(Likely related: #428/#429/#430 are failing on `Nat66Objects`/`EdgenatRuleKey` undefined in the eBPF nat66/edgeprog bindings, which looks like the same DSR/Maglev/NAT66 redesign landing mid-migration across this stack of branches.)

scotwells (approved):

PR #439, opened 2026-08-21
fix: Require manual rollout for BGP router DaemonSets

privateip:
## Summary

fabric-router and galactic-router are BGP speakers whose health checks only confirm the process is running, not that the BGP session has reconverged. Kubernetes was free to restart their pods automatically on any image or spec change, advancing to the next node before the previous one's session actually recovered. For galactic-router's route-reflector role, a restart drops every peer's session at once instead of just one node's. Both DaemonSets now require a manual pod delete to apply an update, and the README documents the rollout procedure and ordering.

## Test plan

- [ ] Applying the updated manifests does not restart any existing fabric-router or galactic-router pod
- [ ] Deleting a single pod picks up the new spec and its BGP session reports Established again afterward


scotwells (approved):
Wonder if readiness probes could be used to control the rollout? Maybe the readiness check could check convergence?

privateip:
> Wonder if readiness probes could be used to control the rollout? Maybe the readiness check could check convergence?

It can but thats not a reliable check.   If the remote peer is not under our administrative control, it could cause sessions to be down and convergence to be partial but in reality our side is both properly configured and healthy

PR #445, opened 2026-08-25
chore: Regenerate NOTICE file

privateip:
## Summary

Third-party license attribution had drifted from go.mod: `go.datum.net/network/api/v1alpha1` and `google.golang.org/grpc` had moved to newer versions without a matching NOTICE update. This regenerates NOTICE from the current dependency set.

## Test plan

- [ ] NOTICE version strings match go.mod


privateip:
superseded by #490

PR #461, opened 2026-08-27
revert: Send Router Advertisement to tap-attached VM guests

privateip:
## Summary

Reverts #460. The Router Advertisement send was a one-shot test stopgap, not a durable fix, and the risk/benefit needs another look before it stays on main.

## Test plan

- [ ] Build, lint, and unit tests pass


scotwells (approved):

drewr (commented):
👍

PR #469, opened 2026-08-28
refactor: Redesign node labels so galactic-cni/router run on edge too

privateip:
## Summary

An SRv6 egress incident traced back to edge nodes never running galactic-cni at all, since the old node-label scheme only let compute nodes opt in to it. This redesigns the labels around two mode-enum keys (galactic and fabric, each router/control) so galactic-cni and galactic-router run on both compute and edge, while galactic-router-rr and the not-yet-implemented fabric-router-rr stay structurally mutually exclusive with the plain mode instead of relying on separate boolean flags someone has to keep in sync. galactic-router is also split out of galactic-gateway's pod into its own independent DaemonSet, so a crash in either no longer risks the other's pod, not just its binary.

## Test plan

- [ ] Every changed Kustomization renders without error and the affinities/taints match the new label scheme
- [ ] A containerlab lab deploy schedules galactic-cni and galactic-router onto both compute and edge nodes, and galactic-router-rr only onto the route-reflector node
- [ ] `task ci` passes clean


ecv (approved):
so, this is good. but holy comments! – consider a repository rule to rely on issues and PRs to explain code, instead of big blobs of comments?

PR #470, opened 2026-08-28
feat: Log BGP peer session transitions in galactic-router

privateip:
## Summary

galactic-router had no way to see BGP peer session transitions in its own logs, since GoBGP's internal logger is wired to a discarded sink regardless of configured level. This adds a dedicated GoBGP peer-event watcher that logs every FSM transition per peer, including why a session dropped when it leaves Established. It runs independently of the periodic BGPPeer status update, which still reflects current state into the CRD but never logged the transition history.

## Test plan

- [ ] Unit tests pass for the FSM state mapping and transition-diff/dedup logic
- [ ] `task lint` and `task build` pass
- [ ] Peer transitions appear in galactic-router logs when a session bounces in a live cluster


scotwells (approved):
We should look at pushing these as events into the k8s control plane

PR #476, opened 2026-08-29
fix: Stop GC from reaping the ingress sidecar's VRF routing state

### LINKED ISSUE #475: A working VPC backend connection can silently stop working minutes later
## Summary

A gateway pod's connection to a VPC backend can fail with a 503 even right after deploying the fix meant to prevent exactly that, with no error anywhere pointing at why. The backend is reachable, its routing state looks fully set up, and restarting the pod does not help.

This surfaced in the staging lab immediately after rolling out the ingress-to-VPC-backend datapath fix: one backend kept failing while others worked, and every symptom pointed at a piece of routing state disappearing moments after it was created, with no log entry marking its removal as intentional.

## Acceptance criteria

- [ ] A gateway pod's connection to a VPC backend stays working indefinitely, not just immediately after the pod starts
- [ ] Any process that removes routing state a sidecar just registered explains, in a log line, why it considered that state orphaned

privateip:
## Summary

The ingress sidecar's own VRF routing state lives under a block reserved so it never collides with a real BGP-backed attachment. The existing GC sweep only recognizes state backed by a live CRD, so it treats every entry the sidecar registers as an orphan and deletes it on its very next tick, moments after the sidecar writes it.

That silently blackholes the sidecar's own outbound connections to VPC backends into a 503, with nothing in the logs pointing at GC as the cause. Confirmed live: one backend's routing entry was gone within a tick of being created, while the state GC never touches stayed intact.

The sweep now also treats the sidecar's reserved block as live unconditionally, the same way a CRD-backed entry already is. A genuinely stale entry outside that block is still removed.

## Test plan

- [ ] Unit tests pass, including a new one covering both the exemption and that ordinary staleness cleanup still works
- [ ] A gateway pod's connection to a VPC backend keeps working well past its first minute up

Fixes #475

### DIFF
--- a/internal/gc/gc.go
+++ b/internal/gc/gc.go
@@ -558,6 +558,31 @@ func SweepEBPFVRFTable(ctx context.Context, k8s client.Client, namespace, nodeNa
 		live[usidmap.VRFKey{Block: block, Argument: argument}] = struct{}{}
 	}
 
+	// internal/ingresssidecar registers its own vrf_table entries under
+	// uformat.BlockMax -- a block deliberately reserved so it can never
+	// collide with a real BGPRouter locator (see that package's
+	// ingressSidecarBlock doc comment) -- and owns their entire lifecycle
+	// itself (ensureEgressDatapath/removeEgressDatapath), independent of
+	// any BGPVRFInstance CRD. Without this, every entry it ever registers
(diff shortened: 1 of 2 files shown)


scotwells (approved):

ecv:
damnit scot i was seconds away

PR #485, opened 2026-08-30
feat(trunk): shared SO_MARK trunk datapath for ingress sidecar

privateip:
internal/ingresssidecar (the second container in the shared Envoy
Gateway fleet's pod) created one synthetic veth pair per VPC it needed
backend connectivity to, dispatching usid_egress by that attachment's
own ifindex. That doesn't scale: interface count grows linearly with
the number of backend VPCs one Envoy pod serves.

Replaces the per-VPC veth with one shared trunk interface per pod,
disambiguated by SO_MARK (skb->mark on egress) instead of by which
interface a packet arrived on:

- usid.c: new mark_vrf_table map (skb->mark -> (Block, Argument)) and
  a usid_egress dispatch fallback -- consulted only when
  ifindex_vrf_table misses, so ordinary per-attachment tenant
  veth/tap traffic is unaffected by construction. Two new drop
  reasons for observability.
- internal/plumbing/ebpf/markvrfmap: new package wrapping
  mark_vrf_table, mirroring ifindexvrfmap's Register/Unregister/
  Get/List/Reconcile/OpenPinned API shape.
- internal/ingresssidecar/ebpfdatapath.go: ensureEgressVeth/
  egressVethNames (one veth pair per VPC, enslaved into that VPC's
  own VRF) are replaced by ensureTrunk (one shared, unenslaved veth
  pair per pod, attached once) plus ensureVRFRoute/removeVRFRoute
  (each VPC's own gateway route into its own table, all naming the
  same shared trunk as nexthop -- the existing "gateway, not
  on-link" route mechanism never required the nexthop device to be a
  member of the table's own VRF). markForTableID derives a
  placeholder mark value from the VPC's own VRF table ID, explicitly
  flagged for network-services-operator's own phase-2 work
  (feat/srv6-routing-identity) to replace.

cmd/galactic-vrf runs one process per pod, in that pod's own network
namespace, so "one trunk per pod" collapses to "one trunk per
process" -- no new pod-identifier-keyed interface naming scheme is
needed; trunkInnerName/trunkPeerName are plain fixed constants.

The trunk and its usid_egress attachment are now process-lifetime,
not per-VPC-lifetime: removeEgressDatapath tears down only the
calling VPC's own route and map entries, never the shared interface
or its attachment.

## Testing

`TestEnsureEgressDatapath_SharedTrunkAcrossVPCs` (rewritten from
`TestEnsureEgressDatapath_AttachesToVethPeerNotVRF`) is this change's
own exit criterion: two VPCs sharing one trunk both resolve
correctly, and tearing one VPC down disturbs neither the trunk nor
the other VPC's still-live state. This also empirically confirms the
one assumption this design rests on that wasn't obvious from Linux
networking theory alone -- an unenslaved trunk inner end still
routes correctly across multiple VRF tables via the peer's
link-local gateway.

Verified locally: `task build`, `task lint`, `task test:unit`, and
`bash scripts/ci.sh unittest-root` (root-gated suite, including the
new markvrfmap package and the rewritten integration test above) all
pass clean.

Scoped entirely to internal/ingresssidecar and the shared
internal/plumbing/ebpf/* packages -- internal/cni and
internal/cnibgp (ordinary tenant CNI ADD/DEL) are untouched; tenant
pods keep today's per-attachment veth path.

## Known gap

No test in this repo sends an actual packet through the compiled
`usid_egress` program with a non-zero `skb->mark` to exercise the new
mark-hit C branch directly (mark-miss is exercised implicitly by
existing tests). Doing so would require a new Go mirror of the
kernel's `__sk_buff` UAPI struct for `cilium/ebpf`'s `Program.Run` --
no such harness exists in this repo today, so it wasn't added as a
side effect of this change.

https://claude.ai/code/session_01CXtDd2iQycCviDx89Tq8to

### DIFF
--- a/internal/ingresssidecar/ebpfdatapath.go
+++ b/internal/ingresssidecar/ebpfdatapath.go
@@ -18,7 +18,7 @@ import (
 
 	"go.datum.net/galactic/internal/plumbing/ebpf/attach"
 	"go.datum.net/galactic/internal/plumbing/ebpf/egressroutemap"
-	"go.datum.net/galactic/internal/plumbing/ebpf/ifindexvrfmap"
+	"go.datum.net/galactic/internal/plumbing/ebpf/markvrfmap"
 	"go.datum.net/galactic/internal/plumbing/ebpf/uformat"
 	"go.datum.net/galactic/internal/plumbing/ebpf/usidmap"
 	"go.datum.net/galactic/internal/plumbing/srv6"
@@ -36,9 +36,9 @@ import (
 var ebpfPinDir = attach.PinDir
 
 // ingressSidecarBlock is the fixed, synthetic uSID Block this file uses for
-// every vrf_table/ifindex_vrf_table entry it registers, standing in for
-// the real, per-router bgp.srv6Locator-derived Block internal/cnibgp's own
-// registerEBPFDatapath uses for a genuine tenant CNI attachment.
+// every vrf_table entry it registers, standing in for the real, per-router
+// bgp.srv6Locator-derived Block internal/cnibgp's own registerEBPFDatapath
+// uses for a genuine tenant CNI attachment.
 //
 // Why a synthetic value is correct here, not a shortcut: vrf_table's key
 // (block<<12 | argument) is consulted by usid_egress purely as a local,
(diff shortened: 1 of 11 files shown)


scotwells (approved):

privateip:
Not implemented

Issue #472, opened 2026-08-28
A gateway pod can permanently lose its route to a reachable VPC member


privateip:
## Summary

After a gateway pod or the node under it restarts, some tenant traffic can lose its path to a VPC member and never get it back, even after another restart. The affected connections just time out, with nothing else about the pod looking wrong. This surfaced in the staging lab, where one tenant's egress path stayed broken for hours while others sharing the same restart recovered within minutes. There is also no operator-visible signal distinguishing "still converging" from "stuck for good."

## Acceptance criteria

- [ ] A gateway pod's route to any genuinely reachable VPC member comes up on its own within a few seconds, without depending on unrelated traffic happening to reach the same address first
- [ ] Restarting the pod is never required to unstick a route that should already work

PR #518, opened 2026-09-04
fix(ingresssidecar): keep a route a second EndpointSlice still claims

scotwells:
## Summary

A gateway pod can permanently lose its path to a healthy VPC backend, serving 503 for a member that every layer above it reports as ready. It happens whenever two EndpointSlices name the same backend address and the older one expires, taking the shared kernel route with it. The surviving slice still records that route as installed, so nothing retries, no error is raised, and the path stays dead until the pod is restarted. This change keeps the route in place while any live claimant remains, and still removes it once the last one goes.

## Why it happens

Routes are tracked per EndpointSlice but removed from the kernel by address, and those two keys are not the same thing. Two slices legitimately name one route while a slice is being republished under a new name, so whichever expires first deletes a route the other is still using. The survivor keeps its installed flag, which makes every subsequent reconcile a no-op — recovery needs a full restart, because only startup seeding rebuilds routes from scratch.

## The fix

Teardown now checks whether another live claimant still holds the same VPC and address before removing anything, and skips the removal when one does. The expired entry is forgotten either way, so nothing leaks, and the route is still removed once its last claimant expires.

## Test plan

- [x] New test reproduces the defect — fails against the current teardown with the route deleted out from under the survivor, passes with the guard
- [x] Companion test confirms the route is still removed once every claimant has expired
- [x] Existing store suite passes under the race detector
- [x] Reproduced live in the staging lab — an affected backend served 503 while its route was absent, and 200 once the route was restored

Local verification ran the store logic in an isolated harness, because this package needs generated eBPF bindings and a Linux kernel to build; CI covers the full suite.

## Follow-up

Nothing re-verifies kernel state against desired state, so a route already lost this way stays lost until the process restarts — that is what happened in staging. Worth closing separately rather than widening this change.

Refs #472. That issue describes this exact symptom and both of its acceptance criteria are violated again today, so it may be worth reopening.

### DIFF
--- a/internal/ingresssidecar/store.go
+++ b/internal/ingresssidecar/store.go
@@ -250,14 +250,24 @@ func (s *Store) Sweep(ctx context.Context, now time.Time) {
 		}
 		if r.installed {
 			v, ok := s.vrfs[r.vpc]
-			if !ok {
+			switch {
+			case !ok:
 				slog.Error("ingresssidecar: sweep found route with no tracked VRF", "key", key, "vpc", r.vpc)
-			} else if err := s.backend.RemoveRoute(r.prefix, v.tableID); err != nil {
-				s.countError("remove_route")
-				slog.Error("ingresssidecar: remove route", "key", key, "vpc", r.vpc, "error", err)
-				liveVPCs[r.vpc] = struct{}{} // keep the VPC alive; retry next sweep
-				pendingRoutes++
-				continue
+			case s.prefixClaimedByOtherLocked(key, r.vpc, r.prefix, now):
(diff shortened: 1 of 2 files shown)


ecv (dismissed):

privateip:
@scotwells can you clean up the lint errors?

scotwells:
@privateip updated!

ecv (approved):

privateip:
One more rebase needed

ecv (changes requested):
@scotwells ugh trunk moves so fast

Issue #293, opened 2026-08-06
docs: Clean up inline comments and transfer to documentation (post-PR #288)


privateip:
## Context

After [PR #288](https://github.com/datum-cloud/galactic/pull/288) (*feat: Grant CNI DaemonSet BPF privileges and wire up e2e*) merges, do a pass through the touched files to:

1. **Remove verbose inline comments** that explain the "why" in code
2. **Transfer that content into documentation** where it belongs (e.g., `docs/cni/configuration.md`, `docs/agent-startup.md`, or new docs as needed)

## Origin

Requested by @ecv in review on PR #288:

> let's do a pass through this to remove comments and xfer them into docs, AFTER it merges?

## Scope

Review all files added/modified in PR #288 for comments that would be better suited as documentation. Target docs location based on topic:

- CNI plugin behavior → `docs/cni/`
- Router / eBPF daemon startup → `docs/router/` or `docs/agent-startup.md`
- Configuration / deployment → existing config docs or `config/` README
- General architecture → `docs/agents/ARCHITECTURE.md` or `docs/agents/CONVENTIONS.md`

## Acceptance Criteria

- Inline comments in changed files are minimal (only non-obvious invariants or workarounds remain)
- Transferred content is integrated into the appropriate doc files
- No loss of information — the "why" is preserved, just moved to docs

Issue #296, opened 2026-08-06
Move eBPF CI/build rationale out of code comments and into documentation


privateip:
## What needs to happen

The eBPF codegen toolchain work in #282 answered several review questions (clang version pinning, why the committed `.o` is embedded rather than regenerated in the container image, why `test-unit-root` re-runs as root, the `-idirafter` multiarch workaround, etc.) by writing the rationale directly into code comments across `.github/workflows/ci.yaml`, `Taskfile.yaml`, `containers/galactic-cni/Dockerfile`, and `internal/plumbing/ebpf/prog/doc.go`. That's fine for getting the PR merged, but it leaves a lot of design reasoning embedded in comments scattered across four files instead of somewhere a reader would think to look for it.

This should move into proper documentation — likely alongside the existing `docs/agents/ARCHITECTURE.md` and `docs/` conventions — so the eBPF toolchain's design decisions are discoverable in one place rather than only inline where the config happens to live.

## Why this matters

Anyone touching CI, the Taskfile, or the Dockerfile later has to piece the reasoning back together from comments spread across unrelated files. Consolidated documentation makes the toolchain's tradeoffs (clang pinning, artifact drift checking, root-gated tests) easier to find and keep current.

## Desired outcome

The design rationale currently living in code comments in `ci.yaml`, `Taskfile.yaml`, the CNI `Dockerfile`, and `doc.go` is captured in a docs page, with the inline comments trimmed back to what's needed to orient a reader at the point of the code itself.

Follow-up from [ecv's review](https://github.com/datum-cloud/galactic/pull/282#pullrequestreview-4878203600) on #282.

Related to #282

privateip:
fixed in #521

PR #521, opened 2026-09-07
docs: Trim function docstrings to purpose and contract

privateip:
## Summary

Comments had grown to about a third of the non-test source, carrying issue numbers, milestone references, reviewer names, and narration of designs that were rejected or already replaced.

That bulk buries the contract a reader actually needs, and six comments had drifted far enough to describe behavior the code no longer has.

Docstrings now state why a function exists, what its inputs and returns mean, and what it does, keeping non-obvious invariants and the reason for each. Comment volume drops by about a quarter and no code changed.

## Test plan

- [ ] Lint, unit tests, and eBPF compilation all pass
- [ ] The diff changes comments only, with no code touched

Related to #293

### DIFF
--- a/cmd/galactic-bgp/main.go
+++ b/cmd/galactic-bgp/main.go
@@ -58,21 +58,21 @@ func newRootCommand() *cobra.Command {
 				return version.All.Encode(os.Stdout)
 			}
 
-			// Real CNI runtimes always pipe the network config JSON on
-			// stdin and close it. If stdin is an interactive terminal
-			// instead, no config will ever arrive and skel's blocking
-			// stdin read would hang forever — print version info instead.
+			// Real CNI runtimes always pipe the network config on stdin and
+			// close it. When stdin is an interactive terminal instead, no
+			// config ever arrives and the library's blocking read would hang
+			// forever, so print version info instead.
 			if term.IsTerminal(int(os.Stdin.Fd())) {
 				fmt.Printf("%s version %s\n", appName, metadata.Version)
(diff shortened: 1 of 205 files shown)


privateip:
Most of this diff is prose churn. The part worth checking is the six comments that described behavior the code does not have, since those were actively misleading rather than merely long.

1. The egress route installer claimed the datapath resolved link and L2 addresses per packet via a FIB lookup. Registration resolves them when the route is written, so an unreachable gateway fails there instead of silently at packet time. The FIB lookup from that attach point unconditionally blackholes, which is why the resolution moved to the control plane in the first place.
2. Two datapath maps were documented as populated once at startup and never per attachment. The control plane rewrites both on every attachment add, idempotently, because there is no once-per-node hook to hang them off.
3. Three comments in the rule reconciler referenced a primary-node assignment function and a status field. Neither exists anywhere in the tree; both went away with the move to anycast, where every gateway node serves every rule identically.
4. The interface resolver claimed a deployment whose NICs leave a VLAN tag in place would need parsing the datapath lacks. The datapath does pop a hardware-offloaded tag from skb metadata. Only a tag left in packet data is still unhandled, which is the narrower claim the comment now makes.
5. A binding reconciler described the writer of its spec as a future extension of a reconciler in this repo. Nothing here constructs that spec, and the writer is external.

Two things I left alone deliberately. Test files are untouched, roughly 4500 comment lines, since their comments are per-case rationale rather than docstrings. The temporary trace counters in the uSID datapath keep their remove-once-resolved note, because that describes present code rather than history.

Related to #296, which asks for the eBPF build rationale to move into documentation. This pass condensed those comments in place and did not move them, so that one stays open.

scotwells (approved):

PR #525, opened 2026-09-11
feat(nat): Add NAT64 egress by generalizing the NAT66 tier into galactic-nat

### LINKED ISSUE #880: Prototype: galactic-nat64 generalized from galactic-nat66 (design from #866)
### Summary

Build a working prototype of the `galactic-nat` design proposed in
[enhancements/networking/nat64-gateway-for-vpc-networks.md](https://github.com/datum-cloud/enhancements/blob/main/enhancements/networking/nat64-gateway-for-vpc-networks.md)
(the design deliverable for #866, merged via #879): generalize the
existing `galactic-nat66` binary and `NAT66Shard` CRD into a combined
`galactic-nat` that adds RFC 6146 stateful NAT64 alongside the existing
NAT66 path, rather than a second, near-duplicate binary. This issue
covers a prototype proving the design out end to end — not the full
Phase 1 production rollout.

### Scope

Prototype the pieces the design doc specifies:

- Rename the `NAT66Shard` CRD to `EgressShard`, carrying a shard SID and
  public address(es) per family (IPv6-only, IPv4-only, or both)
- One session table inside the combined binary with structurally
  separate NAT64 and NAT66 entries — distinct table entries and
  translation functions, not a unified code path — keyed by VRFID (the
  SRv6 Argument) and address family
- Shared port allocator serving both families
- Per-VRF eBPF map tracking session count + configured limit, checked
  synchronously at flow-creation time; at-limit behavior is fail-closed
  (drop the triggering SYN/first packet), never evicting an established
  session
- NAT64 traffic always takes ordinary IPv6-in-IPv6 SRv6 encapsulation
  (never the dual-stack IPv4 IPIP path)
- Anti-spoofing / trust validation on shard ingress (new — closes the
  gap NAT66 currently ships without)

### Non-goals for this prototype

- Anycast SID advertisement or regional session-state (#791's Phase 2 —
  explicitly deferred, not part of this prototype)
- Any metrics/session-data emission to telemetry, quota, or insights, or
  any tenant-facing surface (deferred to a later design per the design
  doc's Non-Goals)
- Production rollout, multi-region or HA validation

### Validation

Run the design doc's [End-to-end validation](https://github.com/datum-cloud/enhancements/blob/main/enhancements/networking/nat64-gateway-for-vpc-networks.md#end-to-end-validation)
steps against the prototype:

1. DNS64 synthesizes an address inside the shared NAT64 prefix for a
   known IPv4-only test name
2. The instance connects using that address; connection completes
   through a `galactic-nat` shard
3. Two tenants with identical inner source addresses on different VRFs
   never collide in the session table or each other's session count
4. A tenant at its configured session limit fails closed without
   affecting any other tenant's sessions or its own established ones
5. NAT66's existing test/regression suite passes unmodified against the
   generalized binary with no IPv4 fields configured

### References

- Design doc: [nat64-gateway-for-vpc-networks.md](https://github.com/datum-cloud/enhancements/blob/main/enhancements/networking/nat64-gateway-for-vpc-networks.md)
- Design/discovery: #866
- Parent enhancement: #791
- As-built documentation follow-up: #877
- DNS64 (related): #792

privateip:
Implements [enhancements#879](https://github.com/datum-cloud/enhancements/pull/879). Depends on [network#22](https://github.com/datum-cloud/network/pull/22) (merged; go.mod points at it).

IPv6-only instances get a path to the IPv4 internet. `galactic-nat66` becomes `galactic-nat`, serving NAT66, NAT64, or both from one session table, one port allocator, and one CRD.

## Where this follows the design, and where it doesn't

Shared, as designed: the session table, port allocator, tenant accounting, counters, CRD, and route installation. `conn_key` gains a family byte so a NAT64 row's IPv4-mapped addresses cannot alias a real IPv6 flow in `::ffff:0:0/96`.

**Split, against the letter of the design:** the two translation cores are separate eBPF programs behind a tail-call dispatcher, not branches in one program. NAT66 is pure PAT — the IPv6 header stays an IPv6 header. NAT64 synthesizes an IPv4 header, computes a header checksum IPv6 has no counterpart for, and takes a checksum delta across a pseudo-header whose *family* changed. Inlining both into one program is a verifier-complexity risk, and the design's own "keep the families structurally separate" requirement makes that arrangement worst. Tail calls give each leaf its own instruction budget.

**Scope reduced:** TCP and UDP only. ICMP translation and fragment handling are absent, so this is not yet a complete RFC 6146 implementation and PMTUD across the translator does not work. Each gap increments its own named drop counter rather than passing silently.

## Test plan

- [x] `task lint` — clean
- [x] `task test:unit`
- [x] Datapath suite as root, all 5 programs verifier-accepted: NAT64 forward and return translate real packets, with both checksums verified against independent full recomputes rather than the datapath's own arithmetic; tenant isolation across families; limit fail-closed; fragment/options/unavailable drops
- [x] **NAT66's existing regression suite passes unmodified** against the generalized binary — the design's validation step 5
- [ ] `task test:e2e` — needs a Kind cluster
- [ ] NAT64 end-to-end — **not possible in containerlab**: the lab's transit mesh is IPv6-only with no IPv4 upstream, so enabling NAT64 there would produce shards that translate correctly and then blackhole, indistinguishable from a broken deployment. The lab stays NAT66-only and `deploy/containerlab/resources/galactic-nat/README.md` records what turning it on would require.


ecv (commented):
Diff exceeds auto-review threshold (7536 lines across 98 files) — flagging for human review.

scotwells (dismissed):

scotwells (dismissed):

privateip:
@scotwells this PR is now ready for final review

ecv (approved):

PR #526, opened 2026-09-11
fix: keep a deleted backend from tearing down a live one's route

scotwells:
Deleting a workload can tear down the network route belonging to a different, running workload that has since been given the same address, black-holing its traffic with nothing to restore it. This already happened once in staging, where the affected workload recovered only incidentally a couple of minutes later.

A backend address reaches the ingress sidecar through two endpoint records that share one kernel route, and tearing down either one removed the route from under the other. The route now stays in place while any record still claims it, and a teardown that finds the route already gone completes instead of retrying forever.

One gap stays open: an address reissued inside the teardown grace window can still collide. Closing it needs a design decision about how a route removal proves it is removing the route it installed, so it is not in this change.

## Testing

A regression test covers the shared-address teardown and fails without this change. This repository does not build on macOS without the generated eBPF sources, so CI is authoritative.

Fixes datum-cloud/infra#5176

### DIFF
--- a/internal/ingresssidecar/ebpfdatapath.go
+++ b/internal/ingresssidecar/ebpfdatapath.go
@@ -404,9 +404,18 @@ func ensureRedirectRoute(prefix *net.IPNet, tableID uint32) error {
 // removeRedirectRoute removes the main-table route ensureRedirectRoute
 // installed for prefix. Deleting by destination and table alone is enough to
 // identify it.
+//
+// An absent route is the outcome a removal asks for, so ESRCH counts as
+// success. Reporting failure instead retries a teardown that can never
+// complete, and keeps the entry alive to collide with the next tenant given
+// this address.
 func removeRedirectRoute(prefix *net.IPNet) error {
-	return netlink.RouteDel(&netlink.Route{
+	err := netlink.RouteDel(&netlink.Route{
 		Dst:   prefix,
 		Table: unix.RT_TABLE_MAIN,
 	})
+	if errors.Is(err, unix.ESRCH) || errors.Is(err, unix.ENOENT) {
+		return nil
+	}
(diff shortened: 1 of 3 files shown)


ecv (commented):
Looks good. Two independent fixes, both sound:

- `removeRedirectRoute` now treats ESRCH/ENOENT as success. Deleting an already-absent route is the outcome a teardown wants, so treating it as failure previously retried forever and left the tracked entry alive to collide with the next tenant reusing that address.
- `Store.Sweep` adds `prefixClaimedElsewhereLocked` so tearing down one EndpointSlice's route entry no longer removes the kernel route when a sibling slice (e.g. a federated copy) still claims the same VPC+prefix. Traced through the liveVPCs bookkeeping — the VRF stays live off the sibling's own entry regardless of map iteration order, so no ordering bug there.

New test (`TestStoreSharedPrefixSurvivesSiblingTeardown`) exercises the exact shared-address scenario and checks both route and VRF survive. Reasonable coverage for the fix.

ecv (approved):
Approving. Both fixes hold up:

- `removeRedirectRoute` treating ESRCH/ENOENT as success stops an unsatisfiable teardown retry, and stops the stale entry surviving to collide with the next tenant assigned that address.
- `prefixClaimedElsewhereLocked` keeps a shared kernel route in place while a sibling EndpointSlice still claims the same VPC and prefix. Checked the `liveVPCs` accounting around the new branch: the sibling entry marks the VPC live from its own iteration, so VRF teardown stays correct independent of map iteration order.

`TestStoreSharedPrefixSurvivesSiblingTeardown` covers the pod-owned plus federated-copy case directly and would fail without the fix.

privateip (approved):

PR #527, opened 2026-09-11
fix(radv): release a failed advertisement attempt's context

scotwells:
The CNI installer runs on every compute node and sets up guest networking there. It leaks memory every time it fails to start an address advertisement for a tap attachment, and it runs under a memory limit tight enough that the leak eventually gets it killed. When it dies it takes node networking setup down with it, including the neighbor resolution that decapsulated traffic needs to reach a guest.

The leak is unbounded in practice rather than theoretical. An attachment whose host interface no longer exists can never start, so it fails and retries every couple of seconds for as long as the attachment record survives. A node in that state today is retrying six of them continuously. A failed attempt now releases what it held.

Two related problems on the same component are real but need a design decision, so they are tracked in datum-cloud/infra#5175 rather than fixed here. Nothing reconciles stale attachment records against the interfaces that exist, and the memory limit is small enough that a single shell command inside the container can breach it.

## Testing

The existing retry test now also asserts that a failed attempt is released. This repository does not build on macOS, because the eBPF bindings are generated at build time and the network namespace dependency excludes darwin, so CI is authoritative.

Related: datum-cloud/infra#5175

### DIFF
--- a/internal/installer/installer.go
+++ b/internal/installer/installer.go
@@ -594,7 +594,16 @@ func reconcileRadvActors(ctx context.Context, actors *radvActorSet) {
 // it, reconcileRadvActors would stay convinced the actor is running. Deleting a
 // key that is already gone, because the attachment itself disappeared, is a
 // safe no-op.
+//
+// Each attempt derives its context from one that lives as long as the process.
+// Only cancelling detaches the derived context. Deleting the entry alone leaves
+// it attached, so a retry loop grows this daemon's memory without bound.
 func radvActorFailed(actors *radvActorSet, iface string) {
+	cancel, ok := actors.cancel[iface]
+	if !ok {
+		return
+	}
+	cancel()
 	delete(actors.cancel, iface)
 }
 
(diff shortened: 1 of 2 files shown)


ecv (commented):
Looks good. Fixes a real leak: `radvActorFailed` deleted the actor's cancel-func entry without invoking it, so a failed advertisement attempt's context stayed derived from the process-lifetime context forever, growing unbounded on repeated retries. The added nil-check on the map lookup avoids a panic if the entry is already gone. Test asserts the cancel func is actually invoked, not just that the map entry clears, which is the right regression guard here.

ecv (approved):
Approving. `radvActorFailed` now cancels the attempt's context before dropping the map entry, closing an unbounded memory growth path where each failed advertisement retry left a context attached to the process-lifetime parent. The map lookup guard keeps a double-call safe. Test asserts the cancel func actually fires rather than only that the entry clears, which is the behavior that regressed.

privateip (approved):

Issue #550, opened 2026-09-14
A shard cannot return a translated reply to a tenant hosted on another node


privateip:
## Summary

After un-translating a reply, a shard looks the tenant's own address up in its node's routing table. The shard does not host that tenant and holds no VRF for it, so the lookup falls through to the management default and the packet is dropped, counted as a missing neighbour. The connection state needed to send it back is present — the flow is tracked — but nothing uses that state to encapsulate the reply toward the node the request came from.

This is the ordinary case, not an edge case: a node never uses its own shard, so every masqueraded flow returns to a node that does not host the tenant. Confirmed by giving the masquerade address a working underlay route first, so the reply reached the shard and still went nowhere — seven drops for a single request, with the next hop resolved and reachable.

## Outcome

A tenant completes a TCP session with a host outside the fabric through its shard.

Sits behind #549, which has to land first for a reply to reach the shard at all.

privateip:
The SRv6 half of this is fixed and measured in #566, which stays a draft until #565 clears. A reply now reaches the tenant's pod and a UDP round trip completes on both families; TCP does not, for a cause #566 makes visible rather than introduces.

Worth correcting the description above, since it sent me the wrong way at first: the shard does re-encapsulate from connection state, and always did. What it could not do was anything useful with the outer source it was handed, because the tenant's node stamped its underlay interface address there.

Issue #549, opened 2026-09-14
A masqueraded egress flow has no return path: the shard's public address never reaches the underlay


privateip:
## Summary

Each shard publishes its masquerade address, but as an EVPN path into the router mesh. The transit carries plain IPv6 unicast and never learns it, so a reply from anything outside the fabric is routed by the transit's default route and discarded.

Verified end to end against a host outside the fabric: a tenant's request arrived there with the correct translated source and port, the host answered, and the transit had no route at all for the address it was answering to. Every masqueraded flow is one-way today. The ingress VIP has the same shape of problem and is worked around by originating a covering aggregate into the underlay from the edge nodes; the shards do nothing equivalent.

## Outcome

A reply from outside the fabric, addressed to a shard's masquerade address, reaches that shard.

Related to #409, which asks where such an address should come from. This is that a correctly configured one is still unreachable.

PR #559, opened 2026-09-15
fix(nat): Give each tenant its own Argument in the egress shard SID

### LINKED ISSUE #538: Egress shard routes carry no per-tenant Argument, so same-node tenants can share a translation
Two tenants on the **same node** with overlapping ULAs can share one connection row in an egress shard's session table. The second tenant's replies are then re-encapsulated toward the first tenant's pod — cross-tenant delivery, not a dropped packet.

Cross-*node* collisions are already closed by keying flows on the encapsulation source (#525). This is the remaining half.

## Cause

`installEgressRoutes` (`internal/cnibgp/bgp.go`) registers the operator-configured shard SID **verbatim**. Nothing writes the attachment's VRFID into the SID's Argument field (bits 69–80), so every tenant VRF encapsulates toward the identical destination and `read_argument` returns the same constant for all of them.

Confirmed on the wire in containerlab — three tenants in three VPCs, one destination address:

```
fd20:10:ff01::100:0  ->  2001:db8:ff02:9:e001::
fd20:20:ff01::100:0  ->  2001:db8:ff02:9:e001::
fd20:30:ff01::100:0  ->  2001:db8:ff02:9:e001::
```

ULAs are RFC 4193 locally assigned and not globally coordinated, so two tenants picking the same prefix is realistic rather than contrived. It is exactly what the Argument was meant to disambiguate.

## Fix, in two parts

**1. Write the Argument at route installation.** `argument uint16` is already a parameter of the enclosing function at the call site, so this part is small.

**2. Widen the shard's BGP advertisement.** `shardAdvertisementPrefixes` (`internal/controller/egressshard_controller.go`) advertises a `/128`. Once each tenant registers a different address, only one has a route, and `table.Register` resolves link and L2 information as part of the write — so every other tenant's CNI ADD fails. The shard must advertise `<block>:<node-id>::/64` covering its Argument space.

Part 2 changes what every node in the fabric learns, so it wants review on its own terms.

## Decision needed first

The two Argument namespaces have different scopes. `allocateArgument` filters `BGPVRFInstances` by `RouterRef.Name`, making Arguments unique **per node**. A shard receives from every node, so it needs values unique across all of them — tenant A on dfw and tenant B on sjc can both hold `0x005` today, and both would write `0x005` into the same shard's locator.

| Option | Cost |
| ------ | ---- |
| Fabric-unique VRFIDs | Capacity drops from 4095 VPCs per node to 4095 per fabric |
| Shard-side allocation | Control-plane round trip on the pod attach path, and a new failure mode there |
| Hash the VPC identifier | Not viable at 12 bits — birthday collisions reach 50% around 75 VPCs |

Either way, 12 bits caps distinguishable tenants per shard locator at 4095.

## Blocked on this

- Per-tenant session limits, removed from the MVP in #525 because they consume a guarantee the fabric does not yet make. Restoring them is mechanical once this lands.
- A same-node variant of `verify:nat-collision`, which today covers only the cross-node case.

## Scope

Not a regression, and not urgent — cross-node was the bulk of the exposure and is closed. Until this lands, isolation is structural for tenants on different nodes and best-effort for tenants sharing one, which is weaker than [enhancements#879](https://github.com/datum-cloud/enhancements/pull/879) claims.

privateip:
## Summary

Two tenants on one node with overlapping ULAs could share a single connection row in an egress shard's session table, so one tenant's replies reached the other.

Nothing wrote an attachment's own identifier into the part of the shard address a shard reads to tell tenants apart, so every tenant VRF on a node encapsulated toward an identical destination.

Each attachment now writes its own, and a shard advertises the address range those values span rather than the one address an operator configured.

> [!IMPORTANT]
> The advertisement has to converge across the fabric before the per-tenant addresses reach any node, or traffic either fails to attach or is discarded at the destination site.

## Test plan

- [x] `task ci` (lint, build, unit, e2e)
- [ ] `task verify:nat-collision-samenode`: two VPCs on one node sharing an address each get their own translation
- [ ] `task verify:nat-collision` and `task verify:nat-datapath` still pass

Fixes #538


privateip:
On the decision #538 said had to come first: it does not. The issue listed three options for making Arguments unique across the fabric, each costly, one of them dropping capacity from 4095 VPCs per node to 4095 per fabric. None is needed. `conn_key` already composes `tenant_arg` with `encap_src`, and `nat.c` says so directly: "Neither alone identifies a tenant fabric-wide. Together they do, which is why both are in the key." Tenant A on dfw and tenant B on sjc holding the same node-local value differ in `encap_src` and get separate rows. Per-node allocation is sufficient, so this is a pure thread-through of a value that was already in scope at the call site.

The scoping argument returns only if something is later keyed on `tenant_arg` alone. Per-tenant session limits are the obvious candidate, and they should be keyed on the same pair rather than the Argument by itself.

The shard datapath needed no change at all. `locator_matches` already compares only the top 64 bits, and its comment says widening it "breaks the intended case of two tenants' egress packets sharing this shard_sid with different Argument values." The XDP side was written for this.

On the advertisement widening, the issue predicted that without it every other tenant's CNI ADD fails, because `Register` resolves the SID through `RouteGet` and rejects a result that does not leave over an SRv6 uplink. That holds where the host route is the only route. It does not hold in containerlab, where each site originates its locator as a /48 into the underlay: a remote node resolves the new address through that aggregate, the ADD succeeds, and the packet is discarded by the aggregate's Null0 at the far site. Silent blackhole rather than a failed attach, and the harder of the two to diagnose. Sequencing the advertisement ahead of the rewrite avoids both, and there is no in-band signal that catches it afterward.

On the lab fixture, ns71 takes its address from `ipam.addresses` rather than a pool because the pool path cannot produce the collision. `internal/cni/ipam` keys allocation state by pool CIDR alone, node-local and not per VPC, so two VPCs configured with the same subnet on one node are handed two different /96s and never collide. The values are byte for byte what the pool would hand the first pod out of that prefix, so nothing downstream of IPAM behaves differently from any other tenant in the lab.

The pass condition is two distinct masquerade ports, matching the cross-node test. That is observable on the forward path alone, which matters because the return leg is still broken: the reply re-encapsulates toward `ResolveNodeSourceAddress`, the node's underlay-facing address. It sits outside the locator block, so it misses `locator_table` at step 2 of `usid_ingress`, and it carries no Argument, so nothing selects a tenant VRF on the way back in. Related to #549 and #550.

scotwells (approved):

Issue #551, opened 2026-09-14
Egress verification passes on a datapath where no session can complete


privateip:
## Summary

The egress check watches for a translated packet on the wire and stops there. It never confirms that packet arrived anywhere, and never confirms a reply came back, so it reports green on a datapath where no egress session can complete.

Both return-path faults found by driving real traffic to a host outside the fabric sat behind a passing check. The lab now has such a host permanently attached to the transit, so a round trip is available to test against.

## Outcome

The egress check drives a request and its response through a shard to a destination outside the fabric, and fails when either direction breaks.

Related to #546, #549 and #550.

privateip:
The egress check now drives a request and its response through a shard to a destination outside the fabric and fails when either direction breaks. This landed as `verify:nat-datapath` in PR #552 ([9541e7d](https://github.com/datum-cloud/galactic/commit/9541e7d67b838d8eebb242b0ae9bd8ddf435ca6)), which added the off-fabric `remote-host` and asserted both directions on both families: the far end must see the translated packet masqueraded to the shard's address, and the tenant's fetch must get a reply back.

The shallow wire-only check (`verify:nat-egress`) remains as the fast check. The return-path faults it used to sit behind are fixed and covered elsewhere: #549 by `verify:nat-return-route`, #550 by the shard stamping the tenant's own End.DT46 SID, and #565 by both legs of the shard transmitting from the driver so conntrack sees the reply. The change is in release v0.2.0.

Closing as already fixed. If the round-trip check still reports green on a broken datapath on the latest version, reopen with the lab state and I will look again.

PR #562, opened 2026-09-15
fix(containerlab): Count only drops the egress check caused

### LINKED ISSUE #546: Egress verification fails on drops that already healed
## Summary

The shard's drop counters only increase, and they live in a pinned map that outlives the process that writes them. One transient — a link flap, a fabric reconverging after a restart — makes the egress verification fail on every later run, including runs where the datapath is provably healthy. Restarting the shard does not clear them; it takes direct map surgery.

The check cannot tell a drop happening now from one that happened hours ago, so once it has failed it stops carrying information, and a real regression behind it goes unnoticed.

## Outcome

The egress check reports drops that occur during the check itself, and a transient that has since healed does not keep it red.

privateip:
## Summary

A NAT shard's drop counters live in a pinned map that outlives the process writing them, so they count everything the node has ever dropped rather than what a given run caused. The egress verification failed on any non-zero reading, so a single transient that had long since healed turned it red on every later run — including runs where the translation provably worked — and clearing it took direct map surgery. The check now samples the counters around its own probe and fails only on ones that moved, the same shape the ingress gateway check already uses. Once a check is permanently red it stops carrying information, which is what made this worth fixing rather than working around.

## Test plan

- [ ] A healed transient in the counters no longer fails the egress check
- [ ] A drop caused during the probe still fails it, naming the reason and the change
- [ ] The full round-trip check reports only drops from its own run
- [ ] Lint passes

Fixes #546


privateip:
Confirmed against the running gvpc lab before changing anything. A baseline run passed. Three pings from a dfw tenant through sjc's shard then bumped the drop counter by 3 — ICMP counts as a drop because the shard is TCP/UDP only by design — and the next run failed on those 3, despite the translation itself succeeding: it got past the EtherType, masquerade port range, and connection table checks and failed only at the end. Deleting the shard pod left the counter at 3, so a restart really does not clear it.

After the change, the same stale 3 no longer fails the run. Driving ICMP through the shard during the probe window does fail it, reporting `nat66_malformed_forward 3 -> 4`, so the check is narrowed rather than defanged. The lab is back where it started, counters zeroed.

Worth being explicit about the tradeoff: a delta cannot see a drop landing between the opening sample and the probe's first packet, or after the closing one. That is inherent to sampling around a probe, and the ingress gateway check accepts the same limit. Failing forever on a transient is not a better alternative.

Out of scope here: the egress check still stops at a translated packet on the wire without confirming delivery or a reply (#551), and the return path is still broken on both families (#549, #550). Those stay open.

ecv (approved):

Issue #565, opened 2026-09-16
A tenant's TCP session through an egress shard dies after the handshake


privateip:
## Summary

A shard's two directions disagree about whether the kernel is in the path. Outbound, it translates and passes the packet up, so netfilter opens a connection-tracking entry; inbound, it translates and transmits from the driver, so netfilter never sees the reply and that entry never leaves SYN_SENT. The tenant's ACK then fails TCP window validation, is marked invalid, and the node's own kube-proxy rule drops it — every shard drop counter stays flat while the session dies. Only the handshake's first packet ever reaches the far end.

Invisible until #549 and #550 landed and replies started arriving. UDP is unaffected: it has no window to fail, and a UDP round trip through a shard completes today on both families.

## Outcome

A tenant completes a TCP session with a host outside the fabric through its shard.

privateip:
Confirmed live in the containerlab lab, dfw's ns10 tenant egressing through sjc's shard to the off-fabric host, with #550's fix in place so replies reach the tenant.

The off-fabric host receives the SYN and nothing else, then retransmits its SYN-ACK until it gives up. On the shard node, `tcpdump -i any` shows the tenant's ACK and GET arriving already correctly translated — right masquerade address, right allocated port — with no matching outbound packet. The kernel takes them and does not forward them.

What it is:

```
# /proc/net/nf_conntrack on the shard node, mid-probe
ipv6 10 tcp 6 48 SYN_SENT src=2001:db8:9966:3::1 dst=2001:db8:1:40::2 sport=53653 dport=80 [UNREPLIED] ...

# ip6tables -t filter -L KUBE-FORWARD -n -v
   64  7888 DROP  all  ::/0  ::/0  ctstate INVALID nfacct-name ct_state_invalid_dropped_pkts
```

That rule's counter moves by ~11 per probe — the ACK plus the GET's retransmissions — while every `galactic_nat_drops_total` reason stays flat, which is what makes this so quiet from the shard's own side. `nf_conntrack_tcp_be_liberal` is 0, so an ACK acknowledging a sequence the kernel never saw is out of window by definition.

The entry stays `[UNREPLIED]` because the SYN-ACK is claimed by XDP at ingress, un-translated, and transmitted straight back out of the driver. Netfilter is never on that path. The forward leg, by contrast, translates in place and passes up into ordinary forwarding, which is what opens the entry in the first place.

Two directions to fix it, and they are genuinely different bets rather than one obvious answer:

[...]

Issue #554, opened 2026-09-14
A backend's reply to an ingress VIP is dropped at the edge node as an invalid connection


privateip:
## Summary

When a backend answers a request sent to an ingress VIP, the reply never reaches the client. The forward half of the connection crosses the edge node inside an encapsulated packet and bypasses that node's connection tracking, so when the plain reply crosses the same node on its way out, tracking holds no record of the flow and marks it invalid. The standard Kubernetes forwarding rules on the node drop anything in that state. Every reply and every retransmit dies at the last hop before the fabric.

## Impact

This is structural rather than a misconfiguration. Direct server return means the two halves of a connection never traverse the same stateful path, so a reply can never match a tracked flow. It surfaces only once packets reach a backend at all, which is why it sits behind the forwarding fault tracked separately.

## Acceptance criteria

- [ ] A backend's reply to an ingress VIP reaches an off-fabric client.
- [ ] The reply survives the edge node without relying on a connection tracking entry that the forward path never creates.
- [ ] The behaviour holds on a node running the stock Kubernetes forwarding rules, with no hand-added exemption.

privateip:
Reproduced by injecting the encapsulated frame a fixed forwarding path would emit, which sidesteps #553 and gets a real request to a real backend.

The compute node decapsulates it, nginx answers, and the reply leaves with the VIP correctly substituted as its source:

```
2001:db8:6060::1.80 > 2001:db8:1:40::2.44450: Flags [S.]
```

That reply arrives on the edge node's site-internal interface and is never forwarded out the uplink. The route is present and resolves over the uplink, so routing is not the blocker.

The drop is the first rule of `KUBE-FORWARD`, `ctstate INVALID -j DROP`. Its counter rose by exactly three for three SYN-ACK retransmits, and `conntrack -L` holds no entry for the flow.

The forward half reached the compute node through XDP, which bypasses netfilter, and it was encapsulated, so tracking never saw the inner flow. Plain forwarding across the same node is unaffected: an ICMP echo over the identical interface pair forwards normally. Only the return half of a load-balanced connection looks invalid.

Confirmed as the sole remaining fault. With a temporary `-s 2001:db8:6060::/48 -j ACCEPT` ahead of that rule, the reply reached the off-fabric client. The rule was removed afterwards and the chain is back to its original state.

[...]

PR #566, opened 2026-09-16
fix(cni): Source a tenant's encapsulated egress from this node's own SID

### LINKED ISSUE #550: A shard cannot return a translated reply to a tenant hosted on another node

privateip:
## Summary

A shard could not return a translated reply to a tenant on another node, which is every tenant. The shard was never at fault: it does re-encapsulate from connection state, toward the outer source the tenant's node stamped, and that source was the node's underlay interface address — a link prefix nobody originates into the fabric and that no node decapsulates on arrival. A tenant's node now stamps its own End.DT46 SID for that tenant's VRF, the same SID the attachment already advertises, so a reply rides the site locator every fabric-router originates and decapsulates like any other SRv6 packet. Fixing this in the underlay alone would have moved the failure rather than removed it.

> [!IMPORTANT]
> Blocked by #565. Egress still does not complete a TCP session — the shard passes its forward leg into the kernel and transmits its return leg from the driver, so conntrack never sees a reply and the tenant's ACK is dropped as invalid. That failure was invisible until replies started arriving, so `verify:nat-datapath` stays out of the verify chain and this should not merge as a claim that tenant egress works.

## Test plan

- [x] `task ci`, plus the BPF tests under root, which skip otherwise
- [x] A tenant's node stamps its own SID per VRF, and two tenants never share one
- [x] The un-translated reply reaches the tenant's pod, confirmed by capture on its node
- [x] A UDP round trip to the off-fabric host completes on NAT66 and NAT64

Fixes #550


privateip:
Kept as a draft because of #565 rather than because anything here is unfinished. This half is done and measured; merging it while egress still cannot carry a TCP session would read as a claim it can.

On the two tests that look like duplicates of each other: one asserts the datapath splices the right Argument into the right base, the other that two attachments on one node never end up with the same source. The second is the one that matters for tenant isolation, since a shard keys a flow's connection state on exactly this value and would otherwise deliver one tenant's replies into another's VRF. Both fail against the pre-fix datapath, which I checked by reverting the splice alone rather than assuming.

The lab evidence, for anyone re-running it. A capture on the tenant's node during a request to the off-fabric host:

```
eth1 Out  2001:db8:ff01:1:e001:: > 2001:db8:ff02:9:e001:: : fd20:10:ff01::100:0.57372 > 2001:db8:1:40::2.80  [S]
eth2 In   2001:db8:ff02:9:e001:: > 2001:db8:ff01:1:e001:: : 2001:db8:1:40::2.80 > fd20:10:ff01::100:0.57372  [S.]
eth1 Out  2001:db8:ff01:1:e001:: > 2001:db8:ff02:9:e001:: : ... [.] ack 1
```

Outbound carries the node's SID base with the tenant's Argument spliced in. Inbound, the shard addresses the un-translated reply to exactly that, and the pod's own stack answers it — so decap, VRF resolution and delivery all work. That leg counted a `fib_no_neigh` drop before this change. A UDP round trip completes on both families, which is the cleanest proof available while #565 holds TCP down.

One thing I did not do, deliberately. The ingress sidecar registers its VRFs under a synthetic Block, so the Argument spliced into its source names no `vrf_table` entry under the node's real one. That costs nothing today: the sidecar installs only per-prefix routes toward a client's own SID, never a shard default, so no shard reflects its source back. It becomes a real bug the day a sidecar VRF gets a NAT66 default, and there is a comment at that call site saying so.

The branch also carries an unrelated lab commit that was already in the working tree: exempting fabric transit from the same kube-proxy invalid-state drop, for #554 and for asymmetric ECMP across dfw's dual-homed compute node. Separate commit, easy to drop if it should go on its own branch.

ecv (approved):

